Issue #10, opened 2016-01-22
Chapter 1.4


ronslow:
Trying
     (defn q \[ ] (find-path (L-harmonic 1.0 1.0) 0.0 1.0 1.57 0.0 3))

Then
    ((q) 0.0) => 1.0
which is correct
but
    ((q) 0.1) 
getting
    IllegalArgumentException No method in multimethod 'mul' for dispatch value: [java.lang.Double [D]  clojure.lang.MultiFn.getFn (MultiFn.java:156)

alexgian:
Erm, why are you defining "q" as a function?

I haven't got the system in front of me to verify, but I suspect it'll work thus:

(def q (find-path (L-harmonic 1.0 1.0) 0.0 1.0 1.57 0.0 3))

(q 0.0) => 1.0
(q 0.1) => 0.955...whatever

I'll check later and verify, my apologies in advance if I'm talking nonsense.

alexgian:
Hm, there may well be a problem, perhaps the boss can comment.
Here's what I put in, as close to the Sussman code as possible without any unnecessary Clojurifications
```
(def length count)
(def append concat)
(def Pi (* 2 (asin 1)))


(defn make-path [t0 q0 t1 q1 qs]
  (let [n (length qs) ts (linear-interpolants t0 t1 n)]
  (Lagrange-interpolation-function
  (append (list q0) qs (list q1))
  (append (list t0) ts (list t1)))))
  
  
(defn parametric-path-action [Lagrangian t0 q0 t1 q1]
  (fn [qs]
    (let [path (make-path t0 q0 t1 q1 qs)]
      (Lagrangian-action Lagrangian path t0 t1))))
      

(comment 
Already Defined
(defn L-harmonic [m k]
  (fn [local]
    (let [q (coordinate local) v (velocity local)]
      (- (* 1/2 m (square v)) (* 1/2 k (square q))))))
)
      
(defn find-path [Lagrangian t0 q0 t1 q1 n]
  (let [initial-qs (linear-interpolants q0 q1 n)
        minimizing-qs (multidimensional-minimize (parametric-path-action Lagrangian t0 q0 t1 q1) initial-qs)]
     (make-path t0 q0 t1 q1 minimizing-qs)))
     
(def q (find-path (L-harmonic 1.0 1.0) 0.0 1.0 (/ Pi 2) 0.0 3))

```
On ```(q 0)``` I get  whole lot of activity for a while, culminationg in this, which is a clojure, not a number...
[...]

PR #1, opened 2015-12-20
Hi     (not really a pull request)

alexgian:
This is not really a pull request!  (though I did spot a typo, LOL)

I just wanted to communicate and say a big Thank You for your efforts.
I will definitely be working with this.

I doubt that I could contribute much at this stage, but perhaps I could come up with some documentation/embelishments, if that would be useful.

Best Regards 
and Wishes for the Festive Season

Alex Gian

PR #11, opened 2016-01-24
SICM - Chapter 1.4

alexgian:
The functions from Ch. 1.4 , Computing Actions, as per #10
I've also added a Pi to env, might as well, move it if you think there's a better place for it.
PS.
Oops, lagrange also requires minimize. I think it's all there now... runs fine here.
PPS.
Ah, I see there is a clash in Travis, since you've already got "make-path" and some others in 
*.math.sicm-ch1-test

I wonder if they should be moved into *.math.mechanics.lagrange, or perhaps even a separate SICMbook directory from where they can be made globally available, they are available out of the box under scmutils.

### DIFF
--- a/src/net/littleredcomputer/math/env.clj
+++ b/src/net/littleredcomputer/math/env.clj
@@ -17,7 +17,7 @@
 ;
 
 (ns net.littleredcomputer.math.env
-  (:refer-clojure :exclude [+ - * / zero? ])  ;; partial]) ;; I'm partial to it
+  (:refer-clojure :exclude [+ - * / zero? partial])
   (:require [net.littleredcomputer.math
              [generic :as g]
              [structure :as s]
@@ -85,4 +85,4 @@
 (def ∂ d/pd)
 (def pd d/pd)
 
-(def Pi (* 2 (asin 1)))
+(def Pi (* 2 (asin 1)))  ;; used in SICM code - albeit as :pi
--- a/src/net/littleredcomputer/math/env.clj
+++ b/src/net/littleredcomputer/math/env.clj
@@ -17,7 +17,7 @@
 ;
 
 (ns net.littleredcomputer.math.env
-  (:refer-clojure :exclude [+ - * / zero? partial])
+  (:refer-clojure :exclude [+ - * / zero? ])  ;; partial]) ;; I'm partial to it
   (:require [net.littleredcomputer.math
              [generic :as g]
              [structure :as s]
@@ -84,3 +84,5 @@
 (def D d/D)
 (def ∂ d/pd)
 (def pd d/pd)
+
+(def Pi (* 2 (asin 1)))
--- a/src/net/littleredcomputer/math/mechanics/lagrange.clj
+++ b/src/net/littleredcomputer/math/mechanics/lagrange.clj
@@ -22,6 +22,7 @@
              [generic :refer :all]
              [structure :refer :all]
              [function :refer :all]]
+            [net.littleredcomputer.math.numerical.minimize :refer :all]
             [net.littleredcomputer.math.numerical.integrate :refer :all]
             [net.littleredcomputer.math.calculus.derivative :refer :all]))
 
--- a/src/net/littleredcomputer/math/mechanics/lagrange.clj
+++ b/src/net/littleredcomputer/math/mechanics/lagrange.clj
@@ -223,3 +223,32 @@
   (fn [[_ [q0 q1] qdot]]  ;; local
     (- (* 1/2 m (square qdot))
        (V q0 q1))))
+
+;; more functions from SICM, ch 1.4  
+
+(defn make-path 
+  "SICM 1st ed.  Ch 1.4 footnote #44 p.22"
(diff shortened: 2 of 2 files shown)


littleredcomputer:
oops, now I see you noticed the clash in the top comment.
Anything that's available OOB in scmutils should be the same here, so no controversy.
If you'd like to be the author of this, integrate the new work with the unit tests by removing the local-to-test implementations, and pull in and test the code the user will get. (For make-path, I like my way of assembling the vector better than GJS's). I can happily do it if you prefer.

Now that this might spread, I think we should do the great renaming to sicmutils, including renaming the repository. When I do, I'l add a CONTRIBUTORS.md and add you to it.

alexgian:
I'm happy for you to do it, if you don't mind.  I'll follow closely, and try to conform to that convention in future pull requests.

alexgian:
By the way, what do you think about having all the code in the book in its own directory?  Say "*.math.SICM" or something.
That way, *.math/mechanics will only keep the functions that are only referenced in the book, but NOT the ones that are actually defined/explained.  Might keep *.math.mechanics from getting too large and unmanageable.  
A fine distinction, but I think think important enough.  Also, it will isolate all the book code into one directory/namespace/whatever making easier for both implementors and students to access.   I'm also thinking of writers of future educational passages using the software, here.  You don't want everyone mingling roughshod with the core implementation.  If you agree, it's probably a good idea to do it now, while it's still early.

alexgian:
Congratulations on the Great Renaming!
It's coming along nicely, is it not?

Incidentally have you ever found a bug in scmutils?
While I was musing over operators, I found this, from SICM ed1 p.497 (Our
Notation)

 (print-expression
  (((* (- D 1) (+ D 1)) (literal-function 'f)) 'x))
(+ (((expt D 2) f ) x) (*  -1 (f x)))      ;; <-- this is what you're supposed
to get

(+ -1 (((expt D 2) f) x))  ;; <- this is what you actually get

It's treating the "1" operator as integer, not as identity!  It must have crept
into later versions... Yours seems OK:
(+ (((expt D 2) f) x) (* -1 (f x)))  ;; <-- sicmutils OK!

I don't know whether you communicate with them on such matters, I know there's
an MIT-Scheme dev-blog out there somewhere.

One of the aspects of scmutils that seems not to be implemented yet, is their
use of "Real".  I haven't looked into it, but (p.500):

(def H (literal-function 'H (-> (UP Real (UP Real Real)(DOWN Real Real))Real)))

CompilerException java.lang.RuntimeException: Unable to resolve symbol: Real in
this context, compiling:(/tmp/form-init2831625156339342613.clj:1:29)

I'd be lying if I said I fully understood the scmutils code there wrt the
uppercases and the "->", but I thought I'd mention it anyway.  Should that be
'Real or something?

littleredcomputer:
Alex, I have gone back and forth on this. To review the status quo: I want to define everything the scmutils distribution defines. Usually, these are identical to what is in the book. Occasionally, the internal definitions have a wider scope or a slightly different implementation than what is shown in the book.

In order to prove that sicmutils can run the code in the book, I have been trying to put every expression evaluated in the book into a unit test. The upside is that every time we `lein test` we prove that we're good enough to run the book up thru ch. 3 or so, with more to come. The bad news is that this code is not in a place where it is easy to get to for the users.

What I'd like to do is take an idea you suggested previously and put all the book code, minus the code that is pre-installed in sicmutils, into a literate document that shows input and output. If that were to be live, I'd need a server somewhere that could eval for us. That is very doable, but it will take time. Thinking along these lines, one thing that is sorely missing is `show-expression`. If I had that, I'd have TeX form, and I have found that Khan Academy's `KaTeX` library for JS TeX rendering to be simply awesome. The master plan is to wire all these things together into some kind of web experience, but it's a way off from here.

alexgian:
> I have been trying to put every expression evaluated in the book into a unit test.
Should be simple enough if the tests only evaluate the "culmination" of examples in the book.
That way they will still test everything, since they are dependent on it, but the the test code will be much shorter.

Eg, no need to test everything separately if eg (as per \#1.4) the simple test succeeds
```
(defn L-harmonic  [m k]
  (fn [[_ q v]]
    (- (* 1/2 m (square v)) (* 1/2 k (square q)))))
(def q (find-path (L-harmonic 1.0 1.0) 0.0 1.0 (/ Pi 2) 0.0 3))
;; tests
(q 'x) ==> (+ 1.
   (* .02848618290127808 (expt x 4))
   (* 2.3223002373016943e-2 (expt x 3))
   (* -.5134492696408387 (expt x 2))
   (* 2.197684197627418e-3 x))
(q 0.0) ==>  1.0 
(q 0.1) ==>  0.9951113473440176
```
Keeping, e.g. L-harmonic (and all subsequent similar) in a seperate "literate" doc would allow the user at a glance to tell what's not provided (and what is), which is not immediately obvious from the book.   OK, they could just evaluate the function and see, but I think it would be much *nicer* the way you suggest.way you suggest.

PS. Are you aware that L-harmonic as is, is unhappy with integer arguments?

PR #17, opened 2016-02-01
Arithmetic operations for Operators

alexgian:
These are the most basic operations, they work OK, but are mainly limited to univariate literate-functions.  You can, for instance, now do this correctly, for literal-functions f and g:
```
[sicmutils.env] > ((D ((* (- D g)(+ D 1)) f)) 'x)
(+
 (* -1 (((expt D 2) f) x) (g x))
 (* -1 (g x) ((D f) x))
 (* -1 ((D f) x) ((D g) x))
 (* -1 ((D g) x) (f x))
 (((expt D 2) f) x)
 (((expt D 3) f) x))
```
Some multivariate literal-functions work, e.g.
```
[sicmutils.env] > (def ff (literal-function 'ff [0 0] 0))
ff
[sicmutils.env] > ((*  (partial 1)  (partial 0)) ff) 'x 'y)
(((partial 1) ((partial 0) ff)) x y)))))
[sicmutils.env] > (((-  D  D) ff) 'x 'y)
(down 0 0)))
```
but, on the whole, they require (mapped) arithmetic operations on structures - as we have discussed.  I'll be looking at those next.

### DIFF
--- a/src/sicmutils/operator.clj
+++ b/src/sicmutils/operator.clj
@@ -79,6 +79,7 @@
 
 ;; multiplication of operators is treated like composition.
 (defn- mul
+  "Multiplication of operators is defined as their composition"
   [o p]
   (Operator. (fn [f] (with-meta
                        #(apply (o (p f)) %&)
@@ -86,6 +87,8 @@
              (v/joint-arity [(v/arity o) (v/arity p)])
              'mul))
 
+;; Do we need to promote the second arg type (Number)
+;; to :sicmutils.expression/numerical-expression?? -- check this ***AG***
 (defmethod g/expt
   [::operator Number]
   [o n]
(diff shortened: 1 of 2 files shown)


alexgian:
Hm, not sure why tests, have failed, everything worked this side.
There was one error, a gcd that took too long, or something, but I seem to remember you mentioning it some time in the dim past.

alexgian:
OK, it's a bit long winded, since every case must be specified exhaustively, it would be nice to generailse some arithmetic operations and commutativity, but *it works*.

After **everything** works, including exponentation on multivariate literal functions, we can look at niceties.

alexgian:
OK, I was looking over the travis results only and it seems to me that it is testing expecting the UNFORMATTED results, whereas my tests run with the results formatted and are accepted thus.

Not quite sure what to do about that.
I *could* write different tests, I suppose, but that would seem wasted effort for no real purpose.

I'd like to hear your suggestions on this.

alexgian:
I have commented out those portions of the test that were causing the problem, so it looks OK now, but we need to address the eventuality of this case appearing again in other test code.

littleredcomputer:
Great! The trick is to pass your test expressions through `(simplify)`, and then compare the results to the s-expressions that result from that. You may wonder why your approach didn't work: it's because different expressions representing equal things don't have to generate the same internal expression content _before simplification_, only after. Since we included the `env` namespace in the test class, it feels like we're running in the REPL environment, but not quite: simplification is the last thing that happens, and it only happens in the REPL wrapper, unless you ask for it (as we will now do). In the _code_ environment, arithmetic etc. on things you build is still possible; after simplification, all the magic has gone away and what you're left with, even though it's an s-expression, has had all the type data removed to get it into that form (this is true of scmutils also). When you evaluate, say, `(up 1 2)`, the result is an instance of the `Struct` class, with all that that entails. When you simplify it, you get the s-expression `'(up 1 2)`, which looks right, so it can be printed, but is not a `Struct`. It is the same thing you would get if you wrote `(list 'up 1 2)`. (Sometimes when I look back at what I write it sounds pedantic. I hope you will forgive me if that is so.)

PR #19, opened 2016-02-06
Operators applied to Structures

alexgian:
I have also added an s/map-struct to sicmutils.structure, to allow the application of the operations on the structures.  We may want to review it later.  Basically it's meant to correspond - approximately - to MIT-Sheme / scmutils vector-map (although that one won't work on down vectors):
```
 > (vector-map square (up 2 3 4))
#|
(up 4 9 16)
|#
```
Unit tests indicate all working as they should be

### DIFF
--- a/src/sicmutils/operator.clj
+++ b/src/sicmutils/operator.clj
@@ -19,7 +19,8 @@
 (ns sicmutils.operator
   (:require [sicmutils
              [value :as v]
-             [generic :as g]])
+             [generic :as g]
+             [structure :as s]])
   (:import (clojure.lang IFn)))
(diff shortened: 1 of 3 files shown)


alexgian:
Do not pull this in, I have discovered problems subsequently.

I think that by allowing g/add and g/sub globally to cater for
[:sicmutils.expression/numerical-expression :sicmutils.structure/structure]
the specification for structures is broken.

Shame that, it needs to be isolated somehow, so that it does what's need in that case, but does not become global behaviour.

I'll leave it in my repo for a while, if you want to look at it.

PR #27, opened 2016-04-13
Add arctangent/atan function

swhalen:
This adds the atan function.

I have simply looked at how asin/acos are implemented and added analogous definitions for atan, so there is a possibility that I have missed something.


littleredcomputer:
Thank you very much for the PR. I'll get to it soon. The "official" scmutils defines one- and two-argument versions of atan, which I have been (up to now) too lazy to sort out. Atan, unlike asin, is defined over the whole real line so we don't have to drop into complex arithmetic when the arg has magnitude > 1 as we do for asin and acos. I probably will accept this and then doctor it up for the 2-arg version, etc., after I have a chance to study the function a bit.

swhalen:
Thank you for the feedback. I've taken the liberty of pushing a commit to this branch that corrects the error you pointed out -- where I erroneously dropped to complex arithmetic.

I can see that implementing the two-argument version of atan might require some more thought, so if I make any attempt at doing that I will refrain from pushing to this branch.

PR #29, opened 2017-04-28
fix test typo (was = with only one argument)

scgilardi:


littleredcomputer:
thank you. (wonder how you spotted it!)

scgilardi:
you're welcome. I should have mentioned that. I use https://github.com/pjstadig/humane-test-output . Flagging those instances is the "bonus feature" mentioned on that page.

PR #35, opened 2020-07-06
Convert pattern modules to cljs

sritchie:
This PR:

- adds `lein-cljsbuild` support to `project.clj`, along with the commands required to execute tests in cljs mode.
- adds `cljs` tests to the travis build
- converts the `pattern` modules to `cljc` files, which compile to both clojure and clojurescript.

I'm still refreshing my memory on best practices here for how to structure a joint cljs/clj project, so let me comment inline with my intent.


.travis.yml
```diff
@@ -2,3 +2,9 @@ language: clojure
 branches:
   only:
     - master
+cache:
+  directories:
+  - $HOME/.m2
```
sritchie, line 7: this will slightly help CI times.

.travis.yml
```diff
@@ -2,3 +2,9 @@ language: clojure
...
+script:
+  - lein test
+  - lein test-cljs
```
sritchie, line 10: boom!

project.clj
```diff
@@ -1,28 +1,32 @@
-;
-; Copyright © 2017 Colin Smith.
```
sritchie, line 2: I can back this out if you like! This sort of change can be obnoxious. I've become obsesses with https://github.com/Malabarba/aggressive-indent-mode for Clojure, and single semicolons are treated as end of line comments... so without this change they all snap to the right. Happy to disable for this project, of course, as the formatting is up to you.

project.clj
```diff
@@ -1,28 +1,32 @@
...
+;; This software is distributed in the hope that it will be useful, but
+;; WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;; General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this code; if not, see <http://www.gnu.org/licenses/>.
+;;
+
+(def cljsbuild '[lein-cljsbuild "1.1.8"
```
sritchie, line 20: I like this trick sometimes for dependencies that need to live in multiple places... that said, it looks like I only use it in one spot below. Happy to move this inline.

project.clj
```diff
@@ -41,8 +45,26 @@
                    :benchmark :benchmark}
   :profiles {:uberjar {:aot :all}
              :travis {:jvm-opts ["-Xmx512M"]}
-             :dev {:dependencies [[org.clojure/test.check "0.9.0"]
-                                  [criterium "0.4.4"]]}
+             :dev {:plugins [~cljsbuild
+                             [lein-doo "0.1.11"]]
+                   :repl-options {:nrepl-middleware
```
sritchie, line 50: These additions (this, `cider/piggieback`) allow you to do `cider-jack-in-cljs` and get a clojurescript repl, provided you have `node` installed.

project.clj
```diff
@@ -41,8 +45,26 @@
...
+                                  [criterium "0.4.4"]
+                                  [cider/piggieback "0.5.0"]
+                                  [lein-doo "0.1.11"]]}
              :test {:jvm-opts ["-Xmx512m"]
                     :dependencies [[org.clojure/test.check "0.9.0"]
-                                   [criterium "0.4.4"]]}})
+                                   [criterium "0.4.4"]]}}
+  :aliases {"test-cljs"
+            ["doo" "node" "main" "once"]}
+  :cljsbuild {:builds
```
sritchie, line 61: this is a basic `cljs` build with "simple" optimizations. As we proceed, we'll want to test that all of the functions still work under aggressive compilation mode. But for now I think this is great.

project.clj
```diff
@@ -41,8 +45,26 @@
...
+                                  [cider.piggieback/wrap-cljs-repl]}
+                   :dependencies [[org.clojure/test.check "0.9.0"]
+                                  [criterium "0.4.4"]
+                                  [cider/piggieback "0.5.0"]
+                                  [lein-doo "0.1.11"]]}
              :test {:jvm-opts ["-Xmx512m"]
                     :dependencies [[org.clojure/test.check "0.9.0"]
-                                   [criterium "0.4.4"]]}})
+                                   [criterium "0.4.4"]]}}
+  :aliases {"test-cljs"
```
sritchie, line 59: `lein-doo` seems to be the preferred way to run `cljs` tests these days: https://github.com/bensu/doo

src/pattern/rule.cljc
```diff
@@ -18,7 +18,7 @@
 ;
 
 (ns pattern.rule
-  (:require [pattern.match :refer :all]))
+  (:require [pattern.match :as m]))
```
sritchie, line 21: clojurescript doesn't allow `:refer :all`, so I had to add prefixes here. This and the treatment of macros will be the most disruptive change.

test/pattern/match_test.cljc
```diff
@@ -18,10 +18,12 @@
 ;
 
 (ns pattern.match-test
-  (:require [clojure.test :refer :all]
-            [pattern.match :refer :all]))
+  (:require #?(:clj  [clojure.test :refer :all]
```
sritchie, line 21: this is the way to fork behavior inside these `cljc` files. We really only care about the macros from `clojure.test`, but the syntax to import macros is different for `cljs`.

test/pattern/match_test.cljc
```diff
@@ -32,62 +34,69 @@
...
+    (is (= [{:x '(a b)} '(c)] ((m/match-var :x) {} '((a b) c) receive))))
 
   (testing "match-var-constraint"
-    (is (= [{:x 6} nil] ((match-var :x integer?) {} '(6) receive)))
-    (is (= nil ((match-var :x integer?) {} '(6.0) receive)))
-    (is (= [{:x 6.0} nil] ((match-var :x float?) {} '(6.0) receive)))
-    (is (= [{:x 6.0} '(7.0)] ((match-var :x float?) {} '(6.0 7.0) receive))))
+    (is (= [{:x 6} nil] ((m/match-var :x integer?) {} '(6) receive)))
+
+    ;; Clojurescript treats floats with no mantissa as integers.
```
sritchie, line 51: this test failed, for the reason I describe here. I added a forked-behavior `expected` statement, then a new test with `6.1` that works as before.

test/pattern/rule_test.clj
```diff
@@ -1,179 +1,30 @@
...
 (ns pattern.rule-test
-  (:require [clojure.test :refer :all]
-            [pattern.rule :refer :all]))
+  (:require [pattern.rule :as r]))
 
-(def ^:private !=> (constantly false))
-
-
-(defmacro ^:private rule-1
+(defmacro rule-1
```
sritchie, line 23: this is a little weird. If you want to use a macro in the SAME namespace, you have to move the macro to a separate `clj` file of the same name, and then include it with a special syntax that you'll see in the `cljc` file.

test/pattern/rule_test.cljc
```diff
@@ -0,0 +1,171 @@
...
+;; WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;; General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this code; if not, see <http://www.gnu.org/licenses/>.
+;;
+
+(ns pattern.rule-test
+  #?(:cljs  (:require-macros [pattern.rule-test :refer [rule-1]]))
```
sritchie, line 21: this is the syntax required to import macros that live in the identical namespace. `clj` tests just get the macro by default, since both files declared the same namespace.

test/pattern/rule_test.cljc
```diff
@@ -0,0 +1,171 @@
...
+             (a (:? x float?) (:? y)) => (c (:? y) (:? x))
+             (* (expt (cos (:? x)) (:? n more-than-two?))) => success
+             (* (expt (tan (:? x)) (:? n #(> % 2)))) => (:? n)
+             (* (expt (sin (:? x)) (:? n #(> % 2)))) => (:? #(- (% 'n) 2))
+             (* (expt (bzz (:? x)) (:? n #(> % 2)))) => (:? (subtract-from 'n -2))
+             (expt (sin (:? x)) (:? n at-least-two?)) => (* (expt (sin (:? x)) (:? #(- (% 'n) 2)))
+                                                            (- 1 (expt (cos (:? x)) 2))))
+          RS (r/rule-simplifier R)]
+      (is (= '(b 4 3) (apply-ruleset R '(a 3 4))))
+      (is (= '(c 4 3.1) (apply-ruleset R '(a 3.1 4))))
```
sritchie, line 159: to get this to pass I had to change `3.0` to `3.1`, for the same reason as the other test's failure; clojurescript treats `3.0` as an integer, not a float.

test/sicmutils/runner.cljs
```diff
@@ -0,0 +1,7 @@
+(ns sicmutils.runner
+  (:require [doo.runner :refer-macros [doo-tests]]
+            [pattern.match-test]
+            [pattern.rule-test]))
+
+(doo-tests 'pattern.match-test
```
sritchie, line 6: this is the `runner` namespace. Any tests that we want to run in `cljs` have to be imported here and passed to `doo-tests`.

project.clj
littleredcomputer: no problem, time to roll with the changes

PR #36, opened 2020-07-07
sicmutils.{rules, value} to cljs

sritchie:
We have our first two `sicmutils` namespaces working in `cljs`! I'm pushing this smaller PR first since from ~here on out the protocol implementations diverge and we have to be a bit more careful in supporting both languages.

I'm reminded again of how much I love Clojure's design. Clojurescript doesn't have `Ratio`, `BigInteger` or `BigDecimal`... but of course we can pick a library like [`exact`](https://github.com/gfredericks/exact) and add implementations for the `sicmutils` protocols in an optional submodule that users can include if they like.


sritchie:
@littleredcomputer , I noticed that TravisCI is running for these pull requests, but not updating the PR itself: https://travis-ci.org/github/littleredcomputer/sicmutils/builds/705738524

If you get a chance, I think that you may need to refresh the Github permissions that travis.org holds.

Alternatively, I can report that I've had great success with the new Github Actions CI system on a different project: https://github.com/google/caliban I'd be happy to do a PR to set that up if you're interested.

sritchie:
I also took on the slightly-more-involved task of porting over the `sicmutils.value` and its tests. I'll add a review inline.

src/sicmutils/value.cljc
```diff
@@ -83,47 +100,98 @@
...
+
+     (defn exposed-arities
+       "When CLJS functions have different arities, the function is represented as a js
+  object with each arity storied under its own key."
+       [f]
+       (let [parse (fn [s]
+                     (when-let [arity (re-find (re-pattern #"invoke\$arity\$\d+") s)]
+                       (js/parseInt (subs arity 13))))
+             arities (->> (map parse (js-keys f))
+                          (concat [(.-cljs$lang$maxFixedArity f)])
```
sritchie, line 155: This logic is a little finicky. You get explicit `arity1` etc for the NON-variadic methods. But then there's this `maxFixedArity` hanging around. 

If you did something like this:

```clojure
(fn
  ([x] (inc x))
  ([x y] (+ x y))
  ([x y z] (* x y z))
  ([x y z & xs] (* x y z)))
```

you'd see `maxFixedArity == 3` and then `1, 2, 3` for the other arities. If you did this:

```clojure
(fn
  ([x] (inc x))
  ([x y] (+ x y))
  ([x y z & xs] (* x y z)))
```

You'd see `maxFixedArity == 3`, but explicit `arityN` implementations for `1, 2`. I concatenate them all and use a set to distinct.

src/sicmutils/value.cljc
```diff
@@ -83,47 +100,98 @@
...
+                          (into #{}))]
+         (if (empty? arities)
+           [(alength f)]
+           (sort arities))))
+
+     (defn js-arity
+       "Returns a data structure indicating the arity of the supplied function."
+       [f]
+       (let [arities (exposed-arities f)]
+         (cond (variadic? f)
```
sritchie, line 166: This tries to follow the same logic you used in `clj`. The only difference is that I attempt to generate `:between` cases... happy to remove those if you like. I added tests that demonstrate what the main differences are.

src/sicmutils/value.cljc
```diff
@@ -83,47 +100,98 @@
...
+                      (first arities)
+                      (last arities)])))))
+
 (def ^:private reflect-on-arity
   "Returns the arity of the function f.
   Computing arities of clojure functions is a bit complicated.
   It involves reflection, so the results are definitely worth
   memoizing."
   (memoize
-   (fn [f]
```
sritchie, line 92: this was moved up, unchanged.

src/sicmutils/value.cljc
```diff
@@ -178,7 +246,7 @@
 (defn ^:private primitive-kind
   [a]
   (cond
-    (or (fn? a) (= (class a) MultiFn)) ::function
```
sritchie, line 181: any reason to keep this the old way? There's no `class` method in `cljs`, so I moved this to the style you were using elsewhere.

test/sicmutils/rules_test.cljc
```diff
@@ -18,26 +18,30 @@
...
     (is (= '(* (sqrt x) (expt x 3)) (s '(expt (sqrt x) 7))))
     (is (= '(expt x 4) (s '(sqrt (expt x 8)))))
     (is (= '(sqrt (expt x 7)) (s '(sqrt (expt x 7)))))))
 
 (deftest divide-numbers-through-test
-  (let [d divide-numbers-through]
-    (is (= 1/2 (d '(/ 1 2))))
+  (let [d r/divide-numbers-through
+        ;; cljs doesn't support ratios.
+        one-half #?(:clj 1/2
```
sritchie, line 35: I'm sure this is going to become a problem. Would love to hear your thoughts on exact arithmetic. Should we take a library dependency for cljs only?

test/sicmutils/value_test.cljc
```diff
@@ -0,0 +1,138 @@
...
+  (is (= [:exactly 0] (v/arity 'x)))
+  (is (= [:at-least 0] (v/arity (constantly 42))))
+  ;; the following is dubious until we attach arity metadata to MultiFns
+  (is (= [:exactly 1] (v/arity [1 2 3])))
+  (let [f (fn [x] (+ x x))
+        g (fn [y] (* y y))]
+    (is (= [:exactly 1] (v/arity (comp f g))))))
+
+#?(:cljs
+   (deftest arities-cljs
```
sritchie, line 49: these are new, and demonstrate what's different. Feel free to tell me to back out the `:between` change!

test/sicmutils/value_test.cljc
```diff
@@ -0,0 +1,138 @@
...
+  (is (= [:exactly 1] (v/arity (fn [x] (+ x 1)))))
+  (is (= [:exactly 2] (v/arity (fn [x y] (+ x y)))))
+  (is (= [:exactly 3] (v/arity (fn [x y z] (* x y z)))))
+  (is (= [:at-least 0] (v/arity (fn [& xs] (reduce + 0 xs)))))
+  (is (= [:at-least 1] (v/arity (fn [x & xs] (+ x (reduce * 1 xs))))))
+  (is (= [:at-least 2] (v/arity (fn [x y & zs] (+ x y (reduce * 1 zs))))))
+  (is (= [:exactly 0] (v/arity 'x)))
+  (is (= [:at-least 0] (v/arity (constantly 42))))
+  ;; the following is dubious until we attach arity metadata to MultiFns
+  (is (= [:exactly 1] (v/arity [1 2 3])))
```
sritchie, line 43: the `cljs` code actually picks up that a vector has arity `[:between 1 2]`, since you can call it or pass a default... we could expose that by calling `(ifn? arg)` instead of `(fn? arg)` in `arity`.

test/sicmutils/value_test.cljc
```diff
@@ -0,0 +1,138 @@
...
+
+(deftest exactness
+  (is (v/exact? 1))
+  (is (v/exact? 4N))
+  (is (not (v/exact? 1.1)))
+  (is (not (v/exact? 'a)))
+  (is (not (v/exact? :a)))
+  (is (not (v/exact? "a"))))
+
+#?(:clj
```
sritchie, line 95: `cljs` doesn't have these data types.

src/sicmutils/value.cljc
```diff
@@ -55,9 +70,11 @@
                                                             (constantly 0)
                                                             {:arity (arity o)
                                                              :from :object-zero-like})
-                       :else (throw (UnsupportedOperationException. (str "zero-like: " o)))))
-  (one-like [o] (cond (number? o) 1
-                      :else (throw (UnsupportedOperationException. (str "one-like: " o)))))
+
+                       :else (unsupported (str "zero-like: " o))))
+  (one-like [o] (if (number? o)
```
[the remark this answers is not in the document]
littleredcomputer: sg

src/sicmutils/value.cljc
```diff
@@ -55,9 +70,11 @@
                                                             (constantly 0)
                                                             {:arity (arity o)
                                                              :from :object-zero-like})
-                       :else (throw (UnsupportedOperationException. (str "zero-like: " o)))))
```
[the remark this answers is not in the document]
littleredcomputer: just fine.

src/sicmutils/value.cljc
```diff
@@ -83,47 +100,98 @@
 ;;   [:between m n]
 ;;   [:at-least m]
 
+#?(:clj
+   (defn jvm-arity [f]
```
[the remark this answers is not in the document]
littleredcomputer: I had forgotten how many rounds in the ring I went with arity! This looks good.

src/sicmutils/value.cljc
```diff
@@ -83,47 +100,98 @@
...
+       "Returns true if the supplied function is variadic, false otherwise."
+       [f]
+       (boolean (.-cljs$lang$maxFixedArity f)))
+
+     (defn exposed-arities
+       "When CLJS functions have different arities, the function is represented as a js
+  object with each arity storied under its own key."
+       [f]
+       (let [parse (fn [s]
+                     (when-let [arity (re-find (re-pattern #"invoke\$arity\$\d+") s)]
```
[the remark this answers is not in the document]
littleredcomputer: It's sad that this requires such a hack in both languages. But thank heaven for IFn, without it we'd be sunk. Lack of a suitable replacement for the MIT-Scheme "make this thing callable" hack has frustrated other efforts to move this codebase

src/sicmutils/value.cljc
littleredcomputer: no reason. this part of the code is beaten until it obeys, and then left alone.

test/sicmutils/rules_test.cljc
littleredcomputer: Good question. There's a subtlety here. Suppose you have an expression like x + y + 1/2. Inside the simplifier, we want to work in Z(x), (orthodox mathematical notation with which you will be familiar) to get cancellation across the vinculum. The simplifier looks to see if there are non exact numbers in the expression, and if so, all bets are off! I thought it would be clever to regard rationals as exact, and so, when the expression above went into the simulator, it would come out as (2x + 2y + 1)  / 2, which is in Z(x) as desired. But I found that, while this worked and enabled some simplifications that MIT overlooked, this slowed everything down, so I reverted it. So, from the point of view of the simplifier, 1/2 and 0.5 are the same thing. I think we can get away without ratio at least for the present. I don't think it will be an impediment to the work we want to do right now.

test/sicmutils/value_test.cljc
littleredcomputer: I need to page-in some context to remember what this was all about. I'm sure what you're doing is fine.

PR #38, opened 2020-07-07
Convert sicmutils.{expressions, generic}

sritchie:
This got tricky! Macros always have something to teach me. This time it was about how to generate the correct runtime behavior from within the compile-time-expanding macro.

### DIFF
--- a/src/sicmutils/expression.cljc
+++ b/src/sicmutils/expression.cljc
@@ -52,7 +52,7 @@
   [expr]
   (cond (instance? Expression expr) (:expression expr)
         (symbol? expr) expr
-        :else (throw (IllegalArgumentException. (str "unknown expression type:" expr)))))
+        :else (v/illegal (str "unknown expression type:" expr))))
 
 (defn variables-in
   "Return the 'variables' (e.g. symbols) found in the expression x,
--- a/src/sicmutils/expression.cljc
+++ b/src/sicmutils/expression.cljc
@@ -41,12 +41,18 @@
 (defn fmap
   "Applies f to the expression part of e and creates from that an Expression otherwise like e."
   [f e]
-  (->> e :expression f (->Expression (:type e))))
+  (->Expression (:type e)
+                (f (:expression e))))
 
 (defn abstract? [^Expression x]
   ;; TODO: GJS also allows for up, down, matrix here. We do not yet have
   ;; abstract structures.
   (= (:type x) ::numerical-expression))
+
+(defn expression?
+  "Returns true if the argument is an expression, false otherwise."
+  [x]
+  (instance? Expression x))
 
 (defn expression-of
   [expr]
--- a/src/sicmutils/generic.cljc
+++ b/src/sicmutils/generic.cljc
@@ -1,36 +1,36 @@
-;
-; Copyright © 2017 Colin Smith.
-; This work is based on the Scmutils system of MIT/GNU Scheme:
-; Copyright © 2002 Massachusetts Institute of Technology
-;
-; This is free software;  you can redistribute it and/or modify
-; it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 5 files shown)


src/sicmutils/expression.cljc
```diff
@@ -41,18 +41,24 @@
 (defn fmap
   "Applies f to the expression part of e and creates from that an Expression otherwise like e."
   [f e]
-  (->> e :expression f (->Expression (:type e))))
+  (->Expression (:type e)
```
sritchie, line 44: Sorry, compulsively rewrote this to be, imo, slightly clearer! I obviously appreciate the magic of the threading operator :)

src/sicmutils/expression.cljc
```diff
@@ -41,18 +41,24 @@
...
 
 (defn abstract? [^Expression x]
   ;; TODO: GJS also allows for up, down, matrix here. We do not yet have
   ;; abstract structures.
   (= (:type x) ::numerical-expression))
 
+(defn expression?
+  "Returns true if the argument is an expression, false otherwise."
+  [x]
+  (instance? Expression x))
```
sritchie, line 55: this saved me a reader conditional inside a file that required it.

src/sicmutils/generic.cljc
```diff
@@ -1,36 +1,36 @@
...
-  (:require [sicmutils
-             [value :as v]
-             [expression :as x]])
-  (:import [sicmutils.expression Expression]
-           (clojure.lang Keyword)))
+  (:refer-clojure :rename {/ core-div
+                           + core-plus
+                           - core-minus
+                           * core-times}
+                  #?@(:cljs [:exclude [/ + - * divide]]))
```
sritchie, line 25: this was a little annoying. I would have assumed that the `:rename` would also exclude, but I was seeing warnings, so here we go.

src/sicmutils/generic.cljc
```diff
@@ -1,36 +1,36 @@
...
+  (:require [sicmutils.value :as v]
+            [sicmutils.expression :as x])
+  #?(:cljs (:require-macros [sicmutils.generic :refer [def-generic-function]])))
```
sritchie, line 28: check this out. A `cljc` file can require itself to get its own macros.

src/sicmutils/generic.cljc
```diff
@@ -48,16 +48,29 @@
       (abstract-number? x)
       (v/numerical? x)))
 
+(defmacro ^:private fork
```
sritchie, line 51: this is black magic, but it works.

src/sicmutils/generic.cljc
```diff
@@ -97,7 +110,7 @@
 (defmulti simplify v/argument-kind)
 
 (defn ^:private bin+ [a b]
-  (cond (and (number? a) (number? b)) (+' a b)
```
sritchie, line 100: there's no arbitrary-precision addition in `cljs`, so I renamed the core one and use that in each of the following functions.

test/sicmutils/expression_test.cljc
```diff
@@ -18,20 +18,22 @@
...
-                 (walk-expression '(+ 3 (* 4 y) x) {'x 5 'y -2} {'+ +})))))
+    (is (= 12 (e/walk-expression '(+ 3 4 x) {'x 5} {'+ +})))
+    (is (= 0 (e/walk-expression '(+ 3 (* 4 y) x) {'x 5 'y -2} {'* * '+ +})))
+    (is (thrown? #?(:clj Exception :cljs js/Error)
+                 (e/walk-expression '(+ 3 (* 4 y) x) {'x 5 'y -2} {'+ +})))))
 
-(deftest foo
-  (testing "foo"
-    (is (= 1 1))))
+(deftest is-expression
```
sritchie, line 36: couple tests of the new fn.

PR #39, opened 2020-07-07
Porting sicmutils.{euclid, numsymb, complex, numbers} [draft]

sritchie:
Staging the next two.


sritchie:
@littleredcomputer interesting! So I thought I was doing the library a favor by adding a bunch of default implementations for numbers... but of course I realized only later what exactly what going on in `numsymb.clj`, where we start attempting to simplify expressions and maintain exactness even at this stage.

So this is not going to help us at all here: https://github.com/littleredcomputer/sicmutils/pull/39/files#diff-3449ff21aa62706fa015dbdbf79e6d09R20

The reason I did this was that the `numbers` tests don't work by themselves, without those multimethod implementations. The learning begins!

PR #43, opened 2020-07-30
Consolidate numeric types, pave the way for more "generic" generics

sritchie:
This PR:

- adds a few (but not all) missing protocol methods to various types around the codebase
- moves the `v/Value` implementation for numbers into `value` itself, so no import's required to get this working
- modifies the `generic.{+, -, /, *} so they don't use `number?` anymore.

This lets the protocol do its work for allowing us to make the code more generic. Ratios and BigDecimal, etc won't return `true` for `number?` in Clojurescript. Complex numbers won't respond in Java, come to think of it. But these types implement `nullity?` and `unity?`, so we should totally be able to use these higher level generic functions and get some nice value out of them.

### DIFF
--- a/src/sicmutils/complex.cljc
+++ b/src/sicmutils/complex.cljc
@@ -58,7 +58,8 @@
   (zero-like [_] ZERO)
   (one-like [_] ONE)
   (freeze [c] (list 'complex (real-part c) (imag-part c)))
-  (exact? [_] false)
+  (exact? [c] (and (v/exact? (real-part c))
+                   (v/exact? (imag-part c))))
   (numerical? [_] true)
   (kind [_] ::complex))
 
--- a/src/sicmutils/generic.cljc
+++ b/src/sicmutils/generic.cljc
@@ -18,11 +18,7 @@
 ;;
 
 (ns sicmutils.generic
-  (:refer-clojure :rename {/ core-div
-                           + core-plus
-                           - core-minus
-                           * core-times}
-                  #?@(:cljs [:exclude [/ + - * divide]]))
+  (:refer-clojure :exclude [/ + - * divide])
   (:require [sicmutils.value :as v]
             [sicmutils.expression :as x])
   #?(:cljs (:require-macros [sicmutils.generic :refer [def-generic-function]])))
@@ -44,8 +40,7 @@
 
 (defn numerical-quantity?
   [x]
-  (or (number? x)
-      (abstract-number? x)
+  (or (abstract-number? x)
       (v/numerical? x)))
 
 (defmacro ^:private fork
@@ -119,8 +114,7 @@
 (defmulti simplify v/argument-kind)
 
 (defn ^:private bin+ [a b]
-  (cond (and (number? a) (number? b)) (#?(:clj +' :cljs core-plus) a b)
-        (v/nullity? a) b
+  (cond (v/nullity? a) b
         (v/nullity? b) a
         :else (add a b)))
 
@@ -128,8 +122,7 @@
   (reduce bin+ 0 args))
 
 (defn ^:private bin- [a b]
-  (cond (and (number? a) (number? b)) (#?(:clj -' :cljs core-minus) a b)
-        (v/nullity? b) a
+  (cond (v/nullity? b) a
         (v/nullity? a) (negate b)
         :else (sub a b)))
 
@@ -139,9 +132,8 @@
         :else (bin- (first args) (reduce bin+ (next args)))))
 
 (defn ^:private bin* [a b]
-  (cond (and (number? a) (number? b)) (#?(:clj *' :cljs core-times) a b)
-        (and (number? a) (v/nullity? a)) (v/zero-like b)
-        (and (number? b) (v/nullity? b)) (v/zero-like a)
+  (cond (and (v/numerical? a) (v/nullity? a)) (v/zero-like b)
+        (and (v/numerical? b) (v/nullity? b)) (v/zero-like a)
         (v/unity? a) b
         (v/unity? b) a
         :else (mul a b)))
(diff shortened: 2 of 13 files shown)


src/sicmutils/complex.cljc
```diff
@@ -59,7 +59,8 @@
   (zero-like [_] ZERO)
   (one-like [_] ONE)
   (freeze [c] (list 'complex (real-part c) (imag-part c)))
-  (exact? [_] false)
+  (exact? [c] (and (v/exact? (real-part c))
```
sritchie, line 62: I noted this in the previous PR, but noting again - this will do what we expect in cljs, since the underlying `js` implementation of complex numbers keeps ints around when possible. But Java's `Complex` type uses floats full stop, so I don't believe (with the current code) we'll ever see `exact?` return true.

src/sicmutils/generic.cljc
```diff
@@ -18,11 +18,7 @@
 ;;
 
 (ns sicmutils.generic
-  (:refer-clojure :rename {/ core-div
-                           + core-plus
-                           - core-minus
-                           * core-times}
-                  #?@(:cljs [:exclude [/ + - * divide]]))
+  (:refer-clojure :exclude [/ + - * divide])
```
sritchie, line 21: with the changes I'm proposing here, these aliases are no longer required, since we delegate down to the `g/` implementations. These should dispatch properly to the functions here.

src/sicmutils/generic.cljc
```diff
@@ -119,17 +114,15 @@
 (defmulti simplify v/argument-kind)
 
 (defn ^:private bin+ [a b]
-  (cond (and (number? a) (number? b)) (#?(:clj +' :cljs core-plus) a b)
```
sritchie, line 122: this is handled correctly by the `else?` branch. Removing this check makes the system more open, as we can customize behavior in a single place (the `g/add`, etc methods)

src/sicmutils/matrix.clj
```diff
@@ -31,6 +31,7 @@
 
 (deftype Matrix [r c ^PersistentVector v]
   v/Value
+  (numerical? [_] false)
```
sritchie, line 34: I do wonder if you've thought, @littleredcomputer , about splitting this protocol up. The fact that we can get away with unimplemented methods makes it feel to me like there may be some fission available. Just a feeling!

src/sicmutils/numbers.cljc
```diff
@@ -35,16 +35,7 @@
 
 #?(:cljs (def ^:private ratio? (constantly false)))
 
-(extend-type #?(:clj Number :cljs number)
```
sritchie, line 38: I moved this over to the protocol definition, since these types are available out of the box, and we want these types to respond to the protocol methods without having to import `sicmutils.numbers`.

src/sicmutils/value.cljc
```diff
@@ -49,23 +49,37 @@
 #?(:cljs
    (def ^:private ratio? (constantly false)))
 
-(extend-type #?(:clj Object :cljs default)
-  Value
-  (nullity? [o] (and (number? o) (core-zero? o)))
+(extend-protocol Value
```
sritchie, line 52: boom!

src/sicmutils/value.cljc
```diff
@@ -76,12 +90,6 @@
                           o)))
   (kind [o] (primitive-kind o)))
 
-(extend-type nil
```
sritchie, line 79: moved this up, into the `extend-protocol` block.

test/sicmutils/complex_test.cljc
```diff
@@ -46,8 +46,18 @@
 
     (is (= '(complex 10.0 0.0) (v/freeze (c/complex 10))))
     (is (v/numerical? (c/complex 10)))
-    (is (not (v/exact? (c/complex 10))))
-    (is (not (v/exact? (c/complex 0 10.1))))))
+
+    (testing "exact?"
+      (is (not (v/exact? (c/complex 0 10.1))))
+
+      ;; cljs is able to maintain exact numbers here.
```
sritchie, line 53: ah, I forgot I'd added "proof".

sritchie:
@littleredcomputer this should be ready for review! and, wow, Github Actions is smoking fast here on the tests.

src/sicmutils/complex.cljc
littleredcomputer: I know of no requirement that any complex number be thought of as exact. (This might be more of a problem if we were to venture into algebraic territory like working with polynomials over Z_p or the Hurwitz ring or something like that. If that happened, then we would need to replace the Java complex API with something where we redid the arithmetic over our generic numbers. Don't worry about it for now.)

src/sicmutils/generic.cljc
littleredcomputer: groovy.

src/sicmutils/numbers.cljc
littleredcomputer: brilliant

test/sicmutils/generic_test.cljc
```diff
@@ -59,59 +59,29 @@
...
+    (is (= #?(:clj Long :cljs js/Number) (v/kind 9)))
+    (is (= #?(:clj Double :cljs js/Number) (v/kind 99.0)))))
 
-  (testing "many"
-    (is (= 0 (g/+)))
-    (is (= 3.14 (g/+ 3.14)))
-    (is (= 7 (g/+ 7)))
-    (is (= 10 (g/+ 1 2 3 4)))
-    (is (= 33 (g/+ 3 4 5 6 7 8)))))
+(deftest generic-plus
```
littleredcomputer, line 67: all this is a palpable improvement

PR #57, opened 2020-08-10
Convert sicmutils.polynomial-gcd to cljs

sritchie:
It works! I owe you a code review here, as this one required some more involved changes.


src/sicmutils/polynomial.cljc
```diff
@@ -498,6 +504,7 @@
 (defmethod g/sub [::polynomial ::polynomial] [a b] (sub a b))
 (defmethod g/exact-divide [::polynomial ::polynomial] [p q] (evenly-divide p q))
 (defmethod g/square [::polynomial] [a] (mul a a))
+(defmethod g/abs [::polynomial] [a] (abs a))
```
sritchie, line 507: any reason not to include `g/gcd` here, and link it to `polynomial-gcd/gcd`?

src/sicmutils/polynomial_gcd.cljc
```diff
@@ -18,18 +18,17 @@
...
+  (:require #?(:cljs [goog.string :refer [format]])
+            [sicmutils.generic :as g]
+            [sicmutils.polynomial :as p]
+            [sicmutils.util :as u]
+            [sicmutils.util.stopwatch :as us]
+            [sicmutils.value :as v]
+            [taoensso.timbre :as log]))
 
-(def ^:dynamic *poly-gcd-time-limit* [1000 TimeUnit/MILLISECONDS])
+(def ^:dynamic *poly-gcd-time-limit* [1000 :millis])
```
sritchie, line 31: I had to build a clojure/clojurescript compatible stopwatch protocol, as you'll see.

src/sicmutils/value.cljc
```diff
@@ -129,16 +130,61 @@
                           o)))
   (kind [o] (primitive-kind o)))
 
+;; Override equiv for numbers.
+(defmulti eq argument-kind)
```
sritchie, line 134: this is something we should talk about. I needed to extend equality between js/Number and other numeric types, so I override `IEquiv` for numbers. But I don't want it to be slow... and if I go full multimethod dispatch everything gets REALLY slow for any integer equality. Not god. Let me know what you think of this solution.

src/sicmutils/value.cljc
```diff
@@ -129,16 +130,61 @@
...
+                          [goog.math.Long js/BigInt u/bigint]
+                          [goog.math.Integer js/BigInt u/bigint]]]
+       (defmethod eq [from to] [l r] (eq (f l) r))
+       (defmethod eq [to from] [l r] (eq l (f r))))
+
+     (extend-protocol IEquiv
+       number
+       (-equiv [this other]
+         (cond (number? other) (identical? this other)
+               (isa? (kind other) ::number) (eq this other)
```
sritchie, line 168: I special cased these two. We could do this below, and explicitly check for the `::number` case. That is a little wonky.

sritchie:
@littleredcomputer , this is ready as well.

src/sicmutils/polynomial.cljc
littleredcomputer: No.

src/sicmutils/value.cljc
littleredcomputer: It really looks like we are tightening things up here. I doubt the Scheme is anywhere near as principled (or maybe it is, since the full numeric tower is implemented natively there). Having this functional in JavaScript means somebody has to unify the behavior across types. So strong medicine was going to be necessary

PR #62, opened 2020-08-14
convert sicmutils.rational-function to cljs

sritchie:
All set! I'll leave a review pointing out some subtleties we want to consider.

### DIFF
--- a/src/sicmutils/rational_function.cljc
+++ b/src/sicmutils/rational_function.cljc
@@ -26,9 +26,10 @@
             [sicmutils.numsymb :as sym]
             [sicmutils.polynomial :as p]
             [sicmutils.polynomial-gcd :as poly]
+            [sicmutils.util :as u]
             [sicmutils.value :as v])
-  (:import [clojure.lang Ratio BigInt]
-           [sicmutils.polynomial Polynomial]))
+  #?(:clj
+     (:import [clojure.lang Ratio BigInt])))
 
 (declare operator-table operators-known)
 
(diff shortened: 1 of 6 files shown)


src/sicmutils/rational_function.cljc
```diff
@@ -39,23 +40,46 @@
...
+       (toString [p] (str u " : " v))
+       (equals [_ b]
+               (and (instance? RationalFunction b)
+                    (and (= arity (.-arity b))
+                         (= u (.-u b))
+                         (= v (.-v b)))))]
+
+      :cljs
+      [Object
+       (toString [p] (str u " : " v))
```
sritchie, line 54: @littleredcomputer , what do you think of this `toString` implementation?

src/sicmutils/rational_function.cljc
```diff
@@ -39,23 +40,46 @@
...
+                   (.toString x)
+                   "\"]"))]))
+
+(defn rational-function? [r]
+  (instance? RationalFunction r))
 
 (defn make
   "Make the fraction of the two polynomials p and q, after dividing
   out their greatest common divisor."
-  [^Polynomial u ^Polynomial v]
```
sritchie, line 53: these type hints weren't buying us anything, so I removed them for now. I DO miss all of this as documentation, btw. In a past life I used https://github.com/plumatic/schema to document everything, like this:

```clojure
(s/defn exception-string :- String
  "Returns a nice, pretty string (with good newlines) of the supplied
  throwable."
  [e :- Throwable]
  (binding [ex/*fonts* nil]
    (ex/format-exception e)))
```

that would be a nice thing to add here.

src/sicmutils/rational_function.cljc
```diff
@@ -67,7 +91,7 @@
         cs (into (into #{} cv) (p/coefficients u))
         integerizing-factor (*
                              (if (< lcv 0) -1 1)
-                             (reduce euclid/lcm 1 (map denominator (filter ratio? cs))))
+                             (reduce euclid/lcm 1 (map u/denominator (filter u/ratio? cs))))
```
sritchie, line 94: this sets us up for this all to work once I get fractions implemented.

src/sicmutils/rational_function.cljc
```diff
@@ -243,63 +265,33 @@
...
-  [c ^RationalFunction r]
-  (make (g/mul c (.-u r)) (.-v r)))
+(defmethod g/sub [::rational-function ::rational-function] [a b] (sub a b))
+(defmethod g/sub [::rational-function ::p/polynomial] [r p] (subp r p))
 
-(defmethod g/mul
-  [::rational-function Double]
-  [^RationalFunction r c]
-  (make (g/mul (.-u r) c) (.-v r)))
+(defmethod g/sub [::rational-function ::v/integral] [r c]
```
sritchie, line 282: this is a harder diff to read, but I basically backed us out to `::v/integral` in many spots, and `::v/number` where you wanted support for doubles. I think some of the implementations are missing. We can fill those in with better tests, I think.

Would you mind viewing the file (outside of the diff) and letting me know if you think we've laid this out like we want?

src/sicmutils/rational_function.cljc
littleredcomputer: SG, ":" for ratios is fine

src/sicmutils/rational_function.cljc
littleredcomputer: This I have done, and I liked it fine

PR #64, opened 2020-08-14
Convert sicmutils.series to cljs

sritchie:
This works! I converted series, and moved a few tests out to areas where they have access to derivatives.


littleredcomputer:
> This works! I converted series, and moved a few tests out to areas where they have access to derivatives.

That is unbelievable. Getting this right was a major headache because the place where this gets used in generating the Lie series was very thinly documented in the text and it wasn't entirely obvious how all this is supposed to work out as you can see from the comments about the layered sum (triangular sum might be a better name). The book seems kind of overly hasty for the example to "just work."

PR #66, opened 2020-08-15
Convert sicmutils.{function, operator, calculus.derivative} to cljs

sritchie:
This PR, tragically, had to swap the partial derivative symbol for the far-more-boring `partial`. I'll have to do a review here, since I was hitting some fairly strange evaluation order issues that I really don't understand. Moving tests around got them fixed, but I remain troubled.

UPDATE: I've figured this out and solved it in #81. Phew!


jupyter/book-examples.ipynb
```diff
@@ -1323,8 +1323,8 @@
     "\n",
     "(tex\n",
     "  (up\n",
-    "    (((∂ 1) (L3-central 'm V)) spherical-state)\n",
-    "    (((∂ 2) (L3-central 'm V)) spherical-state)))"
+    "    (((partial 1) (L3-central 'm V)) spherical-state)\n",
```
sritchie, line 1326: hopefully this is not too painful to see!

src/sicmutils/analyze.cljc
```diff
@@ -137,5 +137,6 @@
 
               :cljs
               (let [i (swap! count inc)
-                    suffix (.padStart (str i) 16 "0")]
+                    suffix (-> (.toString i 16)
```
sritchie, line 140: this was a bug. `(.toString i 16)` generates hex, which matches the Clojure behavior. Before I was generating decimal.

src/sicmutils/operator.cljc
```diff
@@ -121,75 +134,62 @@
...
-;; Do we need to promote the second arg type (Number)
+;; Do we need to promote the second arg type (::v/integral)
 ;; to ::x/numerical-expression?? -- check this ***AG***
-(defmethod g/expt
-  [::operator Number]
-  [o n]
-  {:pre [(integer? n)
-         (not (neg? n))]}
+(defmethod g/expt [::operator ::v/integral] [o n]
+  {:pre [(not (g/negative? n))]}
```
sritchie, line 146: I removed the `integer?` precondition and narrowed the dispatch.

src/sicmutils/value.cljc
```diff
@@ -53,6 +54,14 @@
 (def ^:private object-name-map (atom {}))
 
 (def numtype ::number)
+(def seqtype #?(:clj Sequential :cljs ::seq))
+
+;; Allows multimethod dispatch to seqs in CLJS.
+#?(:cljs
+   (do
+     (derive IndexedSeq ::seq)
```
sritchie, line 62: this is annoying! You can dispatch to `Sequential` in `clj` and `isa?` returns true; but Clojurescript is all protocols. So if I'm a type that implements a protocol, that does NOT mean that I respond true to `isa?`. All multimethod inheritance has to happen via these inheritance declarations. I think these are the only ones we need...

test/sicmutils/analyzer_test.cljc
```diff
@@ -25,7 +25,7 @@
 
 (deftest symbol-generator
   (let [gen (a/monotonic-symbol-generator "cake")
-        symbols (repeatedly 200 gen)]
+        symbols (repeatedly 1000 gen)]
```
sritchie, line 28: Figured I'd get the test a little more coverage of possible hexes.

test/sicmutils/runner.cljs
```diff
@@ -1,38 +1,44 @@
...
-            [sicmutils.numbers-test]
-            [sicmutils.numsymb-test]
-            [sicmutils.polynomial-test]
-            [sicmutils.rational-function-test]
-            [sicmutils.rules-test]
-            [sicmutils.series-test]
-            [sicmutils.simplify-test]
-            [sicmutils.structure-test]
-            [sicmutils.value-test]))
+            pattern.match-test
```
sritchie, line 3: Here I just removed the extraneous surrounding vector.

jupyter/book-examples.ipynb
littleredcomputer: No :)

PR #68, opened 2020-08-16
Convert sicmutils.numerical.compile to cljs

sritchie:
Works for common subexpression elimination. To get all tests, going, we'll need to modify the build to make `eval` available. Compiling doesn't work unless you run in self-hosted-mode.


test/sicmutils/numerical/compile_test.cljc
```diff
@@ -18,33 +18,38 @@
...
+            [sicmutils.numerical.compile :as c]))
 
 (def ^:private near (v/within 1e-6))
 
-(deftest compile-univariate
-  (let [f (fn [x] (+ 1 (square (sin x))))
-        cf (compile-univariate-function f)]
-    (is (near (f 0.5) (cf 0.5)))))
+#?(:clj
+   (deftest compile-univariate
```
sritchie, line 33: the actual `eval`-calling functions won't work until we figure out how to run the tests using self-hosted clojurescript.

PR #74, opened 2020-08-16
Convert sicmutils.infix to cljs

sritchie:
Works! But the JS test reveals again that we have slightly different behavior between the simplifier in cljs and clojure.


src/sicmutils/infix.cljc
```diff
@@ -21,15 +21,23 @@
...
             [sicmutils.expression :as x]
-            [sicmutils.numerical.compile :as compile])
-  (:import (java.io StringWriter)))
+            [sicmutils.util :as u]
+            [sicmutils.numerical.compile :as compile]))
 
 (defn ^:private make-symbol-generator
   [p]
   (let [i (atom 0)]
-    (fn [] (symbol (format "%s%04x" p (swap! i inc))))))
```
sritchie, line 32: format is not great in Clojurescript.

src/sicmutils/infix.cljc
```diff
@@ -130,20 +138,18 @@
...
-                                           (str " " op " "))]
-                         (loop [a args]
-                           (.write w (str (first a)))
-                           (if-let [a' (next a)]
-                             (do
+                       (let [sep (case op
+                                   * (or juxtapose-multiply " * ")
+                                   expt "^"
+                                   (str " " op " "))]
+                         (with-out-str
```
sritchie, line 145: `with-out-str` is a nice trick, except if you println in any of the functions called internally. We can go back to an explicit `StringWriter`, if you like and if we want that ability.

src/sicmutils/infix.cljc
```diff
@@ -175,7 +181,8 @@
 
 (defn ^:private digit->int
   [^Character d]
-  (Character/digit d 10))
+  #?(:clj (Character/digit d 10)
+     :cljs (js/parseInt d)))
```
sritchie, line 185: there's no character in cljs, just strings.

test/sicmutils/infix_test.cljc
```diff
@@ -224,22 +230,43 @@
...
                (s->infix expr)))
 
-        (is (= (str "function(dx, dy, f, partial, x, y) {\n"
-                    "  var _0003 = partial(0);\n"
-                    "  var _0004 = [x, y];\n"
-                    "  var _0005 = partial(1);\n"
-                    "  var _0006 = _0005(f);\n"
-                    "  var _0007 = _0003(f);\n"
-                    "  return 1/2 * Math.pow(dx, 2) * _0003(_0007)(_0004) + dx * dy * _0005(_0007)(_0004) + 1/2 * Math.pow(dy, 2) * _0005(_0006)(_0004) + dx * _0007(_0004) + dy * _0006(_0004) + f(_0004);\n}")
+        ;; TODO there is some difference happening internally in the simplifier,
```
sritchie, line 243: frustrating... correct, but strange. We'll come back to this, of course.

PR #81, opened 2020-08-17
[bugfix] Fix overflow issues!

sritchie:
YES! I've discovered the cause of the stack size issues. I'll comment inline about the problem. This fixes all of my issues, I believe, with test ordering, and more complex cases failing. Exciting!


src/sicmutils/analyze.cljc
```diff
@@ -99,8 +99,12 @@
                           (add-symbols! w)
                           (analyze w)))
                       (add-symbols! simplified-expr))))
+
                 (add-symbols! [expr]
-                  (->> expr (map add-symbol!) add-symbol!))
+                  (add-symbol!
+                   ;; FORCE the side effect of binding all symbols.
```
sritchie, line 105: this was the issue. I think the problem is that the lazy calls that would populate the `var->expr` map weren't completed before this check:

https://github.com/littleredcomputer/sicmutils/pull/81/files#diff-60cf0d352bd3ef15f0f72fad2dfea362R122

Or something like that. Any call to `@var->expr` in the body of `backsubstitute` was causing the overflow issue.

```
                        (symbol? expr) (if-let [w (@var->expr expr)]
                                         (backsubstitute w)
                                         expr)
```

src/sicmutils/simplify.cljc
```diff
@@ -87,7 +87,7 @@
 
 (defn ^:private simplify-and-canonicalize
   [rule-simplify canonicalize]
-  (fn simplify [expression]
```
sritchie, line 90: name wasn't getting used

test/sicmutils/polynomial_test.cljc
```diff
@@ -419,6 +419,12 @@
            (A '(* y (sin y) (cos (+ 1 (sin y) (sin y) (expt (sin y) 4)))))))
     (is (= '(+ (expt cos 2) (expt sin 2)) (A '(+ (expt cos 2) (expt sin 2)))))
     (is (= '(+ (expt cos 2) (expt sin 2)) (A '(+ (expt sin 2) (expt cos 2)))))
+    (is (= '(+ (up (+ (/ d a) (/ (* -1 b) a))) (* -1 (up (/ (* -1 y) (+ a b)))))
```
sritchie, line 422: this is a smaller test that catches the problem explicitly.

PR #83, opened 2020-08-17
[proposal] split Value protocol into 5 protocols

sritchie:
@littleredcomputer , idea for you. What if we were to split `Value` up into:

- `INumerical`
- `IExact`
- `IAdditive` or something similar, for `nullity?` and `zero-like`
- `IMultiplicative` or something similar, for `unity?` and `one-like`
- `Value` for `kind` and `freeze`.

The goal here would be to allow for nice default implementations for types that can only satisfy some subset of these.

Of course, something like the ability to declare semigroup, monoid, ring, field etc would be great...

I haven't pushed this through to all the other types that extend `Value`, but wanted to get your thoughts.

Cheers!

### DIFF
--- a/src/sicmutils/value.cljc
+++ b/src/sicmutils/value.cljc
@@ -30,13 +30,21 @@
      (:import (clojure.lang BigInt PersistentVector RestFn Sequential MultiFn Keyword Symbol)
               (java.lang.reflect Method))))
 
-(defprotocol Value
-  (numerical? [this])
+(defprotocol INumerical
+  (numerical? [this]))
+
+(defprotocol IExact
+  (exact? [this]))
+
+(defprotocol IAdditive
   (nullity? [this])
+  (zero-like [this]))
+
+(defprotocol IMultiplicative
   (unity? [this])
-  (zero-like [this])
-  (one-like [this])
-  (exact? [this])
+  (one-like [this]))
+
+(defprotocol Value
   (freeze [this]
     "Freezing an expression means removing wrappers and other metadata from
   subexpressions, so that the result is basically a pure S-expression with the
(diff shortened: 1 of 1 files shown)


src/sicmutils/value.cljc
```diff
@@ -97,58 +105,85 @@
        (derive goog.math.Integer ::integral)
        (derive goog.math.Long ::integral)))
 
-(extend-protocol Value
-  #?(:clj Number :cljs number)
+(extend-type Symbol
```
sritchie, line 108: Here's an example of an override that we took care of, before, with conditionals inside the default impl.

src/sicmutils/value.cljc
```diff
@@ -97,58 +105,85 @@
...
+(extend-type #?(:clj Object :cljs default)
+  INumerical
+  (numerical? [_] false)
+
+  IExact
+  (exact? [x] false)
+
+  IAdditive
+  (nullity? [x] false)
+  (zero-like [_] (if (or (fn? o) (instance? MultiFn o))
```
sritchie, line 122: I could pull this out of the default and extend it to functions and multifn...

littleredcomputer:
Sometimes I wonder why these things can't be handled by defmulti. I know I considered that in the past and backed down. I need to dive back in (as you've been patiently waiting for).

At least as far as protocols are concerned, I don't disapprove, but maybe we should catch breath. I have long thought that there'd be a way to do without freeze; as the code has matured it seems less and less of a thing

sritchie:
@littleredcomputer yes, I totally agree that we need to catch our breath before making changes like this. This is more of a sketch-in-code than something I want to merge! We should have a discussion soon about the project and talk about directions, once we settle down.

I did just have to use `v/freeze` to good effect in #84 to get ratios working across languages, so this is a good time to be thinking about what it might become.

Thanks for the reviews!

littleredcomputer:
BTW I will undertake the work of replacing integrate and minimize in JS,
the first using ODEX in a 1-dimensional mode, for instance. I am hopeful
that its adaptive nature might make it useful for improper integrals.

sritchie:
@littleredcomputer , excellent, can't wait! You'll need this dep:

```
[cljsjs/odex "2.0.4-0"]
```

And then this example snippet should work:

```clojure
(ns example.core
  (:require [clojure.test :refer [deftest is]]
            [odex :as o]))

(defn solve-test []
  (let [s (o/Solver. 1)]
    (.solve s routine 0 #js [1] 1)))

(deftest solver
  (is (< (- 2.7182817799042955
            (first (.-y (solve-test))))
         1e-5)))
```

Notice the `#js [1]` form, which converts `[1]` into the required mutable JS array.

PR #84, opened 2020-08-17
Add fraction.js, sicmutils.ratio for cross-platform ratio support [in progress]

sritchie:
It's time. I'll be able to get to this in the next couple of days...

EDIT: This could be a much better ratio literal for cljs than the mess I have now. https://github.com/mfikes/precise/blob/master/src/precise/tagged_literals.cljc#L17

### DIFF
--- a/src/sicmutils/ratio.cljc
+++ b/src/sicmutils/ratio.cljc
@@ -142,7 +142,7 @@
        (toString [r] (str (v/freeze r)))
 
        IPrintWithWriter
-       (-pr-writer [x writer _]
+       (-pr-writer [x writer opts]
          (let [x (v/freeze x)]
            (if (number? x)
              (write-all writer x)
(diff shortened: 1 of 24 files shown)


src/sicmutils/ratio.cljc
```diff
@@ -0,0 +1,223 @@
...
+     (defmethod op [Ratio ::v/integral] [a b] (f a b))
+     (defmethod op [::v/integral Ratio] [a b] (f a b)))
+
+   :cljs
+   (do
+     ;; The -equiv implementation handles equality with any number, so flip the
+     ;; arguments around and invoke equiv.
+     (defmethod v/eq [::v/number Fraction] [l r] (= r l))
+
+     (defn promote [x]
```
littleredcomputer, line 166: I can't figure out if it's a promotion or a demotion :)

src/sicmutils/ratio.cljc
```diff
@@ -0,0 +1,223 @@
+;;
+;; Copyright © 2017 Colin Smith.
```
littleredcomputer, line 2: Maybe we should replace this with your name, or "the sicmutils authors" the way they do in golang

sritchie: I'm fine with this, or "sicmutils authors", either way!

sritchie:
Closing in favor of #94 , #90 , #89 , #95 and #96 .

PR #85, opened 2020-08-18
Convert sicmutils.env to cljs

sritchie:
Okay, this is an interesting one. I was able to ALIAS everything into `sicmutils.env`... but we don't have a way to do the equivalent of `use` or `:refer :all` in a clojurescript file.

It is possible to use `require` to pull everything into a repl session - just not a compiled cljs file.

I wrote called `bootstrap-repl!` that does this. The examples and tests won't look as good, since we'll have to prefix things with `e` or explicitly call them out for imports...

But if you're `clj` only the old stuff will of course work.


src/sicmutils/env.cljc
```diff
@@ -82,11 +96,12 @@
...
     (if (and as
              (or (sequential? a) m?)
-             (every? integer? as))
+             (every? v/integral? as))
       (if m?
         (matrix/get-in a as)
         (get-in a as))
-      (apply core-ref a as))))
+      #?(:clj (apply core-ref a as)
+         :cljs (get-in a as)))))
```
sritchie, line 104: I just drop back to `get-in` here instead of throwing an exception. What do you think?

src/sicmutils/examples/pendulum.cljc
```diff
@@ -0,0 +1,42 @@
...
+;; You should have received a copy of the GNU General Public License
+;; along with this code; if not, see <http://www.gnu.org/licenses/>.
+;;
+
+;; The simple pendulum supported at the point x (in the plane; x should
+;; be a function of t returning an up-tuple (up xt yt).
+
+(ns sicmutils.examples.pendulum
+  (:refer-clojure :exclude [+ - * / zero? partial ref])
+  (:require [sicmutils.env :as e :refer [D sin cos square + - * / ref]]))
```
sritchie, line 25: not so bad!

src/sicmutils/simplify.cljc
```diff
@@ -272,9 +272,13 @@
 
 (defn expression->stream
   "Renders an expression through the simplifier and onto the stream."
-  [expr stream]
-  (-> (g/simplify expr)
-      (pp/write :stream stream)))
+  ([expr stream]
+   (-> (g/simplify expr)
+       (pp/write :stream stream)))
+  ([expr stream options]
```
sritchie, line 278: It's possible to use this with `cider` by calling

```
(setq cider-print-fn “sicmutils.generic/expression->stream”)
```

I'll document this at some point, but thought I'd prep it for now.

test/sicmutils/env_test.cljc
```diff
@@ -0,0 +1,108 @@
...
+;; ; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;; ; General Public License for more details.
+;; ;
+;; ; You should have received a copy of the GNU General Public License
+;; ; along with this code; if not, see <http://www.gnu.org/licenses/>.
+;; ;
+
+(ns sicmutils.env-test
+  (:refer-clojure :exclude [+ - * / zero? partial ref])
+  (:require [clojure.test :refer [is deftest testing]]
```
sritchie, line 22: I can't believe I didn't know that this form works in cljs and clj... all my crazy test namespaces from before, with the reader conditional, look so goofy now.

littleredcomputer: That feels like a bit of luck :)

PR #95, opened 2020-08-26
Added algebraic laws, #sicm/bigint literal, tests for all numeric types (3)

sritchie:
This PR:

- adds a reader literal for `#sicm/bigint` parsing at the repl and in source
- adds `test.check` generators for many of the types we use.
- adds test.check tests for semigroup, monoid, group, ring and field, with a few variations on each. These are nice gut checks of the `g/add`, `g/mul`, `g/sub`, `g/negate` and `g/invert` implementations.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -23,6 +23,7 @@
 (defproject net.littleredcomputer/sicmutils "0.12.2-SNAPSHOT"
   :description "A port of the Scmutils computer algebra/mechanics system to Clojure"
   :url "http://github.com/littleredcomputer/sicmutils"
+  :scm {:name "git" :url "https://github.com/littleredcomputer/sicmutils"}
   :license {:name "GPLv3"
             :url "http://www.opensource.org/licenses/GPL-3.0"}
   :dependencies [[org.clojure/clojure "1.10.1" :scope "provided"]
--- a/project.clj
+++ b/project.clj
@@ -55,7 +55,9 @@
                                                   #"sicmutils.simplify"]}
                    :repl-options {:nrepl-middleware
                                   [cider.piggieback/wrap-cljs-repl]}
-                   :dependencies [[org.clojure/test.check "1.0.0"]
+                   :dependencies [[org.clojure/test.check "1.1.0"]
+                                  [com.gfredericks/test.chuck "0.2.10"]
+                                  [same/ish "0.1.4"]
                                   [criterium "0.4.5"]
                                   [cider/piggieback "0.5.0"]
                                   [lein-doo "0.1.11"]]}
--- a/src/data_readers.cljc
+++ b/src/data_readers.cljc
@@ -1,0 +1,1 @@
+{sicm/complex sicmutils.complex/parse-complex}
--- a/src/data_readers.cljc
+++ b/src/data_readers.cljc
@@ -1,1 +1,2 @@
-{sicm/complex sicmutils.complex/parse-complex}
+{sicm/complex sicmutils.complex/parse-complex
+ sicm/bigint  sicmutils.util/parse-bigint}
--- a/src/sicmutils/euclid.cljc
+++ b/src/sicmutils/euclid.cljc
@@ -50,6 +50,5 @@
 ;; multimethod implementation for basic numeric types.
 (defmethod g/gcd :default [a b] (gcd a b))
 
-(defn lcm
-  [a b]
-  (g/abs (g/divide (g/* a b) (gcd a b))))
+(defmethod g/lcm :default [a b]
+  (g/abs (g/divide (g/* a b) (g/gcd a b))))
--- a/src/sicmutils/generic.cljc
+++ b/src/sicmutils/generic.cljc
@@ -61,13 +61,20 @@
   "Defines a multifn using the provided symbol. Arranges for the multifn
   to answer the :arity message, reporting either [:exactly a] or
   [:between a b], according to the arguments given."
-  [f a & b]
-  (let [arity (if b `[:between ~a ~@b] [:exactly a])
-        docstring (str "generic " f)
-        klass (fork :clj clojure.lang.Keyword :cljs 'cljs.core/Keyword)]
+  {:arglists '([name arities docstring? attr-map? & options])}
+  [f arities & options]
+  (let [[a b] (if (vector? arities) arities [arities])
+        arity     (if b [:between a b] [:exactly a])
+        docstring (if (string? (first options))
+                    (str "generic " f ".\n\n" (first options))
+                    (str "generic " f ))
+        options (if (string? (first options))
+                  (next options)
+                  options)
+        kwd-klass (fork :clj clojure.lang.Keyword :cljs 'cljs.core/Keyword)]
     `(do
-       (defmulti ~f ~docstring v/argument-kind)
-       (defmethod ~f [~klass] [k#]
+       (defmulti ~f ~docstring v/argument-kind ~@options)
+       (defmethod ~f [~kwd-klass] [k#]
          ({:arity ~arity :name '~f} k#)))))
 
 ;; Numeric functions.
@@ -76,7 +83,12 @@
 (def-generic-function sub 2)
 (def-generic-function div 2)
 (def-generic-function negate 1)
-(def-generic-function negative? 1)
+(def-generic-function negative? 1
+  "Returns true if the argument `a` is less than `(v/zero-like a), false
+  otherwise. The default implementation depends on a proper Comparable
+  implementation on the type.`")
+(defmethod negative? :default [a] (< a (v/zero-like a)))
+
 (def-generic-function exp 1)
 (def-generic-function log 1)
 (def-generic-function abs 1)
@@ -95,6 +107,7 @@
 
 (def-generic-function expt 2)
 (def-generic-function gcd 2)
+(def-generic-function lcm 2)
 (def-generic-function exact-divide 2)
 
 (def-generic-function square 1)
@@ -109,7 +122,7 @@
 (def-generic-function tan 1)
 (def-generic-function asin 1)
 (def-generic-function acos 1)
-(def-generic-function atan 1 2)
+(def-generic-function atan [1 2])
 
 ;; Operations on structures
 (def-generic-function invert 1)
--- a/src/sicmutils/rational_function.cljc
+++ b/src/sicmutils/rational_function.cljc
@@ -85,13 +85,17 @@
   ;; a generic operation (along with lcm), and binding the euclid implmentation
   ;; in for language supported integral types. Perhaps also generalizing ratio?
   ;; and denominator. TODO.
+  ;;
+  ;; (note from sritchie: I don't think this is true anymore, as of 8.26.20.
+  ;; Should we remove the comment, or is there some more work to do to make this
+  ;; more efficient?)
   (let [arity (.-arity u)
         cv (p/coefficients v)
         lcv (last cv)
         cs (into (into #{} cv) (p/coefficients u))
         integerizing-factor (g/*
                              (if (< lcv 0) -1 1)
-                             (reduce euclid/lcm 1 (map u/denominator (filter u/ratio? cs))))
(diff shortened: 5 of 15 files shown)


project.clj
```diff
@@ -55,7 +55,9 @@
                                                   #"sicmutils.simplify"]}
                    :repl-options {:nrepl-middleware
                                   [cider.piggieback/wrap-cljs-repl]}
-                   :dependencies [[org.clojure/test.check "1.0.0"]
+                   :dependencies [[org.clojure/test.check "1.1.0"]
+                                  [com.gfredericks/test.chuck "0.2.10"]
```
sritchie, line 59: these are a couple of libraries that make test.check nicer to use.

src/data_readers.cljc
```diff
@@ -1 +1,2 @@
-{sicm/complex sicmutils.complex/parse-complex}
+{sicm/complex sicmutils.complex/parse-complex
+ sicm/bigint  sicmutils.util/parse-bigint}
```
sritchie, line 2: another literal, this time for `bigint`.

src/sicmutils/euclid.cljc
```diff
@@ -50,6 +50,5 @@
 ;; multimethod implementation for basic numeric types.
 (defmethod g/gcd :default [a b] (gcd a b))
 
-(defn lcm
-  [a b]
-  (g/abs (g/divide (g/* a b) (gcd a b))))
+(defmethod g/lcm :default [a b]
```
sritchie, line 53: `ratio` has a native `lcm`, so this is a nice setup...

src/sicmutils/generic.cljc
```diff
@@ -61,13 +61,20 @@
   "Defines a multifn using the provided symbol. Arranges for the multifn
   to answer the :arity message, reporting either [:exactly a] or
   [:between a b], according to the arguments given."
-  [f a & b]
-  (let [arity (if b `[:between ~a ~@b] [:exactly a])
-        docstring (str "generic " f)
-        klass (fork :clj clojure.lang.Keyword :cljs 'cljs.core/Keyword)]
+  {:arglists '([name arities docstring? attr-map? & options])}
```
sritchie, line 64: I massaged this macro to be a little closer to `defmulti`, so that I could pass in docstrings. wdyt?

src/sicmutils/generic.cljc
```diff
@@ -76,7 +83,12 @@
 (def-generic-function sub 2)
 (def-generic-function div 2)
 (def-generic-function negate 1)
-(def-generic-function negative? 1)
+(def-generic-function negative? 1
```
sritchie, line 86: here's an example.

src/sicmutils/generic.cljc
```diff
@@ -109,7 +122,7 @@
 (def-generic-function tan 1)
 (def-generic-function asin 1)
 (def-generic-function acos 1)
-(def-generic-function atan 1 2)
+(def-generic-function atan [1 2])
```
sritchie, line 125: it does require a vector for multiple arities...

src/sicmutils/rational_function.cljc
```diff
@@ -85,13 +85,17 @@
   ;; a generic operation (along with lcm), and binding the euclid implmentation
   ;; in for language supported integral types. Perhaps also generalizing ratio?
   ;; and denominator. TODO.
+  ;;
+  ;; (note from sritchie: I don't think this is true anymore, as of 8.26.20.
+  ;; Should we remove the comment, or is there some more work to do to make this
+  ;; more efficient?)
```
sritchie, line 91: flagging this for you @littleredcomputer: do we want to remove the comment, or is there more to do here?

src/sicmutils/value.cljc
```diff
@@ -174,8 +174,8 @@
        (js*  "~{} == ~{}" l r))
 
      (doseq [[from to f] [[goog.math.Long goog.math.Integer u/int]
-                          [::exact-integral goog.math.Integer u/int]
-                          [::exact-integral goog.math.Long u/long]
+                          [::native-integral goog.math.Integer u/int]
```
sritchie, line 177: whoops, typo!

src/sicmutils/value.cljc
```diff
@@ -206,6 +200,21 @@
...
+   (extend-type js/BigInt
+     IEquiv
+     (-equiv [this other]
+       (if (= js/BigInt (type other))
+         (js*  "~{} == ~{}" this other)
+         (eq this other)))
+
+     IPrintWithWriter
+     (-pr-writer [x writer opts]
+       (let [rep (if (<= x (.-MAX_SAFE_INTEGER js/Number))
```
sritchie, line 213: would definitely love your thoughts here @littleredcomputer . I'm currently printing bigint as a normal number in cljs, if it's smaller than the max value. That means that we'll be able to get nice output in cases where the numbers aren't too big... but is this a problem? If someone puts a bigint, they probably want to be able to round trip it and make sure that the bigint stays. Thoughts?

test/sicmutils/complex_test.cljc
```diff
@@ -39,6 +43,12 @@
            (read-string {:readers {'sicm/complex c/parse-complex}}
                         (pr-str #sicm/complex "1 + 2i"))))))
 
+(deftest complex-laws
+  ;; Complex numbers form a field. We use a custom comparator to control some
+  ;; precision loss.
+  (with-comparator (sc/compare-ulp 1.0 1e3)
```
sritchie, line 49: Complex field tests, one line... so good!

test/sicmutils/euclid_test.cljc
```diff
@@ -63,15 +63,17 @@
...
+    (is (= 30 (g/lcm 6 15)))
+    (is (= 12 (reduce g/lcm [2 3 4])))
+    (is (= 30 (reduce g/lcm [2 3 5]))))
 
-  ;; CLJS can't currently handle the precision of these numbers.
-  #?(:clj
-     (testing "high precision gcd"
-       (is (= (ok 37279462087332 366983722766) 564958))
-       (is (= (ok 4323874085395 586898689868986900219865) 85)))))
+  (testing "high precision gcd"
```
sritchie, line 73: we can activate this if we use generic operations above. And here's that cross platform reader literal in action.

test/sicmutils/generators.cljc
```diff
@@ -1,2 +1,70 @@
...
+                               :NaN? false
+                               :min excluded-upper
+                               :max max})])))
+
+(def complex
+  (gen/let [r (reasonable-double)
+            i (reasonable-double)]
+    (c/complex r i)))
+
+(extend-protocol si/Approximate
```
sritchie, line 59: this is the protocol that the `same` library uses to compare items. This is nice because it lets use use something like `near`, but nested deep inside arbitrary data structures.

test/sicmutils/laws.cljc
```diff
@@ -0,0 +1,263 @@
...
+             b generator
+             c generator]
+            (is (ish? (g/mul a (g/add b c))
+                      (g/add (g/mul a b) (g/mul a c))))
+            (is (ish? (g/mul (g/add b c) a)
+                      (g/add (g/mul b a) (g/mul c a))))))
+
+;; These put the above laws together into known abstract-algebraic bundles.
+
+(defn additive-semigroup
```
sritchie, line 126: You'll love this, @littleredcomputer

test/sicmutils/numbers_test.cljc
```diff
@@ -84,14 +105,13 @@
...
-     (gt/integral-tests #(BigInteger/valueOf %))))
+     (gt/integral-tests u/biginteger))
 
-#?(:clj
-   (deftest double-generics
-     (gt/integral-tests double
-                        :eq near
-                        :exclusions #{:exact-divide :gcd :remainder :modulo :quotient :negative?})
-     (gt/floating-point-tests double :eq near)))
+(deftest double-generics
```
sritchie, line 110: these work now, not sure why I had siloed them.

src/sicmutils/value.cljc
```diff
@@ -174,12 +174,12 @@
...
      (doseq [[from to f] [[goog.math.Long goog.math.Integer u/int]
-                          [::exact-integral goog.math.Integer u/int]
-                          [::exact-integral goog.math.Long u/long]
+                          [::native-integral goog.math.Integer u/int]
+                          [::native-integral goog.math.Long u/long]
                           [goog.math.Long js/BigInt u/bigint]
                           [goog.math.Integer js/BigInt u/bigint]]]
-       (defmethod eq [from to] [l r] (eq (f l) r))
-       (defmethod eq [to from] [l r] (eq l (f r))))
+       (defmethod eq [from to] [l r] (= (f l) r))
```
sritchie, line 181: this is an important error that the tests exposed. The idea here is that any type mismatch gets delegated to `eq`... but then, when the mismatch is corrected, `eq` needs to delegate BACK to the `-equiv` implementations that each type provides between identical types.

Without this, we were falling back to the default `identical?` check above between numbers, which was failing.

test/sicmutils/generic_test.cljc
```diff
@@ -65,25 +68,43 @@
 
 (deftest generic-plus
   (is (= 0 (g/+)) "no args returns additive identity")
-  (is (= 3.14 (g/+ 3.14)) "single arg should return itself")
-  (is (= 10 (g/+ 0 10 0.0 0 0)) "multi-arg works, as long as zeros appear.")
-  (is (= "happy" (g/+ 0 "happy" 0.0 0 0)) "returns really anything."))
+  (checking "g/+"
```
sritchie, line 71: I converted some of these to `test.check` style.

test/sicmutils/numbers_test.cljc
```diff
@@ -19,15 +19,38 @@
...
+     ;; There's no biginteger / bigint distinction in cljs.
+     (l/field 100 sg/biginteger "java.math.BigInteger"))
+
+  #?(:cljs
+     ;; this is covered by sg/long in clj.
+     (l/ring 100 sg/native-integral "integral js/Number")))
+
+(deftest floating-point-laws
+  ;; Doubles form a field too.
+  (with-comparator (v/within 1e-3)
```
sritchie, line 50: I had to tune the precision here to get this all working... the distributive test puts some pressure on the normal comparison when the numbers get enormous.

test/sicmutils/numbers_test.cljc
```diff
@@ -19,15 +19,38 @@
...
+            [sicmutils.generators :as sg]
+            [sicmutils.laws :as l]
+            [sicmutils.numbers :as n]
+            [sicmutils.value :as v]))
 
 (def near (v/within 1e-12))
 
+(deftest numeric-laws
+  ;; All native types available on clj and cljs form fields.
+  (l/ring 100 sg/bigint #?(:clj "clojure.lang.BigInt" :cljs "js/BigInt"))
```
sritchie, line 36: these use `l/ring` for now, since we can't convert integrals to rationals and get them passing the field tests.

So this is not technically a test of "`BigInt` forms a field"... but more, "the thing you get when you call `u/bigint` is a field, because it can promote itself up the numeric tower."

src/sicmutils/generic.cljc
littleredcomputer: sweet.

test/sicmutils/laws.cljc
littleredcomputer: Yes I do! Hats off 🎩

test/sicmutils/numbers_test.cljc
littleredcomputer: That is really good to note. It's thought provoking to think of how it would be if we were stricter, more haskell-ish, about type, and it really feels like It would be difficult to do. Auto-raising-and-lowering are needed to get the text to work (and to maintain sanity) and as we have noticed the uses of 0 and 1 are slightly less than principled in some places.

Do we use bigint outside of polynomials... If we couldn't form the quotient of bigints, I wonder what would stop working

PR #102, opened 2020-09-02
pure Clojure implementation of Nelder-Mead

littleredcomputer:
Normally I'd have done this with mutable arrays of doubles and all that, but this time I just decided to relax and let a little garbage be generated. I compared traced runs vs. scipy for two examples, so I think it's working.


sritchie:
Fantastic, reviewing now!

sritchie:
Hey @littleredcomputer , this is awesome. I took a pass over it to try and understand the algorithm, and ended up with this:

https://gist.github.com/sritchie/7538314e7effd9adcd0efba2a0e1cb9f

This is a pretty significant workover, so I apologize if this is more intrusive than the normal code review! But I think it does manage to make the algorithm clearer than the original scipy.

I'll make some comments in the PR now, but please take a look.

src/sicmutils/numerical/minimize.cljc
```diff
@@ -36,13 +36,8 @@
               (org.apache.commons.math3.optim
```
sritchie, line 36: I think `MultivariateFunction` above can go too

src/sicmutils/numerical/minimize.cljc
```diff
@@ -90,45 +85,117 @@
   ([f a b]
    (minimize f a b nil)))
 
+(defn ^:private make-initial-simplex
```
sritchie, line 88: I took a crack at how I'd write these functions, mostly because you noted that you're feeling rusty, so I figured I'd see what I could come up with, and see if it was different at all. Gives us something to stare at, in any case.

The main change was using `map-indexed` to get the indices in:

```clojure
(defn ^:private initial-simplex
  "Takes an n-vector x0 and returns a list of n+1 n-vectors, of which x0 is the
  first, and the remainder are formed by perturbing each coordinate in turn."
  [x0 {:keys [nonzero-delta zero-delta]
       :or {nonzero-delta 0.05
            zero-delta 0.00025}}]
  (let [scale (inc nonzero-delta)
        f (fn [i xi]
            (let [perturbed (if (zero? xi)
                              zero-delta
                              (* scale xi))]
              (assoc x0 i perturbed)))]
    (into [x0] (map-indexed f x0))))
```

I ended up making the arguments configurable. Curious to hear what you think here.

src/sicmutils/numerical/minimize.cljc
```diff
@@ -90,45 +85,117 @@
...
+  [x0]
+  (let [nonzdelt 0.05
+        zdelt 0.00025
+        N (count x0)]
+    (into [x0]
+          (for [i (range N)
+                :let [xi (nth x0 i)]]
+            (assoc x0 i (if (zero? xi) zdelt (* (+ 1 nonzdelt) xi)))))))
+
+(defn ^:private sup-norm
```
sritchie, line 100: I would suggest modifying this slightly, so that you can reuse it when you calculate the max norm of the function values themselves. Something like:

```clojure
(defn ^:private sup-norm
  "Returns the absolute value of the distance of the individual coordinate in any
  simplex farthest from its corresponding point in x0."
  [[x0 :as simplex]]
  (let [coords (if (sequential? x0)
                 (mapcat #(v- % x0) simplex)
                 (map #(- % x0) simplex))]
    (reduce max (map u/compute-abs coords))))
```

src/sicmutils/numerical/minimize.cljc
```diff
@@ -90,45 +85,117 @@
...
-  of X and f(X) at each search step."
-  ([f q observe]
-   #?(:cljs
-      (u/unsupported "multidimensional-minimize isn't yet implemented in Clojurescript.")
+  If :callback is supplied, will be invoked with the intermediate points of evaluation.
+  If :info is true, wraps the result with evaluation information.
+  TODO: document other options
 
-      :clj
-      (let [total-time (us/stopwatch :started? true)
```
sritchie, line 102: I'm so happy that you rebuilt this in Clojure! It's much clearer now, and I'm really happy that this will work in Clojurescript. So awesome.

src/sicmutils/numerical/minimize.cljc
```diff
@@ -90,45 +85,117 @@
...
+      Implementing the Nelder-Mead simplex algorithm with adaptive
+      parameters. 2012. Computational Optimization and Applications.
+      51:1, pp. 259-277
+
+  I gratefully acknowledge the Python implementation in SciPy, which I have imitated here.
+  "
+  [func x0 & {:keys [adaptive callback maxiter maxfun xatol fatol info]
+              :or {xatol 1e-4, fatol 1e-4}}]
+
+  (let [v+ #(mapv + %1 %2)                ;; add two vectors elementwise
```
sritchie, line 122: I think for the smaller vectors we have here, we're almost certainly fine... Clojure vectors are pretty quick.

If we want to go faster, the clearest place right now is the place where we swap out the worst element. From my gist:

```clojure
(fn [elem f-elem]
  (let [s  (conj (pop simplex) elem)
        fs (conj (pop f-simplex) f-elem)]
    (sort-by-f s fs dimension)))
```

Which uses this function:

```clojure
(defn ^:private sort-by-f
  "Returns the two inputs `simplex` and `f(simplex)` sorted in ascending order by
  function value.

  Dimension must == the length of each element in the simplex."
  ([simplex f-simplex]
   (sort-by-f simplex f-simplex (count (peek simplex))))
  ([simplex f-simplex dimension]
   (let [indices-by-f (sort-by (partial nth f-simplex)
                               (range 0 (inc dimension)))
         sorted-simplex  (mapv simplex indices-by-f)
         sorted-fsimplex (mapv f-simplex indices-by-f)]
     [sorted-simplex sorted-fsimplex])))
```

Instead of replacing the final element with `(conj (pop ..) ...)`, we could

- make `simplex` and `f-simplex` transients
- write a function that can insert the elements into the correct location...

but as I type it, I'm thinking that this is maybe not worth the complexity. Still, interesting!

src/sicmutils/numerical/minimize.cljc
```diff
@@ -90,45 +85,117 @@
...
+        N (count x0)
+        rho 1.0
+        chi (if adaptive (+ 1.0 (/ 2.0 N)) 2)
+        psi (if adaptive (- 0.75 (/ (* 2.0 N))) 0.5)
+        sigma (if adaptive (- 1 (/ N)) 0.5)
+        maxiter (or maxiter (* N 200))
+        maxfun (or maxfun (* N 200))
+        shrink (fn [s] (let [s0 (first s)]
+                         (into [s0] (map #(v+ s0 (v* sigma (v- % s0))) (next s)))))]
+    (loop [simplex (make-initial-simplex x0)
```
sritchie, line 139: the paper was clearer than the scipy implementation! https://www.researchgate.net/publication/225691623_Implementing_the_Nelder-Mead_simplex_algorithm_with_adaptive_parameters

test/sicmutils/numerical/minimize_test.cljc
```diff
@@ -25,28 +25,46 @@
...
+            (fn [[x0 x1 x2]]
+              (let [log_pdot [(+ x0 x1 x2)
+                              (+ x0 x1)
+                              (+ x0 x2)
+                              x0
+                              x0]
+                    logZ (Math/log (reduce + (map #(Math/exp %) log_pdot)))]
+                (- logZ (+ x0 (* 0.3 x1) (* 0.5 x2)))))
+            [0 0 0]
+            :xatol 1e-5
```
sritchie, line 66: Same here, `:simplex-tolerance`

test/sicmutils/numerical/minimize_test.cljc
```diff
@@ -25,28 +25,46 @@
...
+            :adaptive false)]
```
sritchie, line 67: Then this would become `:adaptive? false`

sritchie (commented):
Oh man, super fun to collaborate on Clojure! Hope you don't mind the extended gist-based implementation.

Here's my take, would love some feedback, and please feel free to take anything from this that you like: https://gist.github.com/sritchie/7538314e7effd9adcd0efba2a0e1cb9f

src/sicmutils/numerical/minimize.cljc
```diff
@@ -90,45 +85,117 @@
...
+      parameters. 2012. Computational Optimization and Applications.
+      51:1, pp. 259-277
+
+  I gratefully acknowledge the Python implementation in SciPy, which I have imitated here.
+  "
+  [func x0 & {:keys [adaptive callback maxiter maxfun xatol fatol info]
+              :or {xatol 1e-4, fatol 1e-4}}]
+
+  (let [v+ #(mapv + %1 %2)                ;; add two vectors elementwise
+        v- #(mapv - %1 %2)                ;; subtract two vectors elementwise
```
sritchie, line 123: I definitely like pulling these out into separate fns, so we can reuse them:

```clojure
(defn- v+
  "add two vectors elementwise."
  [l r]
  (mapv + l r))

(defn- v-
  "subtract two vectors elementwise."
  [l r]
  (mapv - l r))

(defn- v*
  "multiply vector v by scalar s."
  [s v]
  (mapv #(* s %) v))
```

src/sicmutils/numerical/minimize.cljc
```diff
@@ -90,45 +85,117 @@
...
+        v* (fn [s v] (mapv #(* s %) v))   ;; multiply vector v by scalar s
+        ;; should we rely on something for the vector math, or keep this dependency-free?
+        fncall_count (atom 0)
```
sritchie, line 126: I like this. Maybe it belongs in some pattern like

```clojure
(defn ^:private counted
  "Takes a function and returns a pair of:

  - an atom that keeps track of fn invocation counts,
  - the instrumented fn"
  [f]
  (let [count (atom 0)]
    [count (fn [x]
             (swap! count inc)
             (f x))]))
```

src/sicmutils/numerical/minimize.cljc
```diff
@@ -90,45 +85,117 @@
...
+        psi (if adaptive (- 0.75 (/ (* 2.0 N))) 0.5)
+        sigma (if adaptive (- 1 (/ N)) 0.5)
+        maxiter (or maxiter (* N 200))
+        maxfun (or maxfun (* N 200))
+        shrink (fn [s] (let [s0 (first s)]
+                         (into [s0] (map #(v+ s0 (v* sigma (v- % s0))) (next s)))))]
+    (loop [simplex (make-initial-simplex x0)
+           fsimplex (mapv f simplex)
+           iterations 0]
+      (let [indices-by-f-value (sort #(compare (nth fsimplex %1) (nth fsimplex %2)) (range 0 (inc N)))
```
sritchie, line 142: I would recommend pulling these lines out into their own function:

```clojure
(defn ^:private sort-by-f
  "Returns the two inputs `simplex` and `f(simplex)` sorted in ascending order by
  function value.

  Dimension must == the length of each element in the simplex."
  ([simplex f-simplex]
   (sort-by-f simplex f-simplex (count (peek simplex))))
  ([simplex f-simplex dimension]
   (let [indices-by-f (sort-by (partial nth f-simplex)
                               (range 0 (inc dimension)))
         sorted-simplex  (mapv simplex indices-by-f)
         sorted-fsimplex (mapv f-simplex indices-by-f)]
     [sorted-simplex sorted-fsimplex])))
```

src/sicmutils/numerical/minimize.cljc
```diff
@@ -90,45 +85,117 @@
...
+            ;; sort simplex and f(simplex) in ascending order of function value
+            simplex (mapv simplex indices-by-f-value)
+            fsimplex (mapv fsimplex indices-by-f-value)]
+
+        (when callback (callback (first simplex)))
+
+        (if (or (> iterations maxiter)
```
sritchie, line 149: Perhaps slightly clearer to make a function here:

```clojure
(defn ^:private stop-fn
  "Returns a function that returns true if stopping conditions are satisfied by
  its supplied iteration count, simplex and evaluated simplex, false otherwise."
  [f-counter dimension {:keys [maxiter maxfun simplex-tolerance fn-tolerance]
                        :or {simplex-tolerance 1e-4
                             fn-tolerance 1e-4}}]
  (let [maxiter (or maxiter (* dimension 200))
        maxfun  (or maxfun (* dimension 200))]
    (fn [iterations simplex f-simplex]
      (or (> iterations maxiter)
          (> @f-counter maxfun)
          (and (<= (sup-norm simplex)   simplex-tolerance)
               (<= (sup-norm f-simplex) fn-tolerance))))))
```

and bind it like:

```
stop?         (stop-fn f-counter dimension opts)
```

and use it like:

```clojure
(if (stop? iteration simplex f-simplex)
        (if info
          {:result s0
           :value f0
           :iterations iteration
           :fncalls @f-counter}
          s0)
        ...)
```

src/sicmutils/numerical/minimize.cljc
```diff
@@ -90,45 +85,117 @@
...
+                     (<= (reduce max (map u/compute-abs (map #(- % (first fsimplex)) fsimplex))) fatol)))
+          ;; RETURN
+          (if info
+            {:result (first simplex)
+             :value (first fsimplex)
+             :iterations iterations
+             :fncalls @fncall_count}
+            (first simplex))
+          ;; CARRY OUT A STEP
+          (let [xbar (v* (/ N) (reduce v+ (butlast simplex)))
```
sritchie, line 161: Here was my long take on how we might make the steps themselves more forthcoming about how they work (full gist here):

```clojure
(defn ^:private step-defaults
  "Generates the options required for a step of Nelder-Mead. :adaptive? controls
  the set of defaults. If true, they're generated using the supplied dimension;
  else, they're static.
  "
  [dimension {:keys [alpha beta sigma gamma adaptive?] :as m}]
  (let [base (if adaptive?
               {:alpha 1.0
                :beta (+ 1.0 (/ 2.0 dimension))
                :gamma (- 0.75 (/ (* 2.0 dimension)))
                :sigma (- 1.0 (/ dimension))}
               {:alpha 1.0
                :beta 2.0
                :gamma 0.5
                :sigma 0.5})]
    (merge base (select-keys m [:alpha :beta :gamma :sigma]))))

(defn ^:private step-fn
  "Returns a function that performs a single step of nelder-mead. The function
  expects a sorted simplex and f-simplex, and returns sorted results - a pair of

  - [simplex, f(simplex)]

  [This Scholarpedia
  page](http://www.scholarpedia.org/article/Nelder-Mead_algorithm) provides a
  nice overview of the algorithm.

  The parameters in opts follow the convention from [Gao and Han's
  paper](https://www.researchgate.net/publication/225691623_Implementing_the_Nelder-Mead_simplex_algorithm_with_adaptive_parameters)
  introducing the adaptive parameter version of Nelder-Mead:

  :alpha - reflection cefficient
  :beta  - expansion coefficient
  :gamma - contraction coefficient
  :sigma - shrink coefficient
  "
  ([f dimension opts]
   (let [{:keys [alpha beta sigma gamma]} (step-defaults dimension opts)]
     (letfn [(centroid-pt [simplex]
               (v* (/ dimension) (reduce v+ (pop simplex))))

             ;; Returns the point generated by reflecting the worst point across
             ;; the centroid of the simplex.
             (reflect [simplex centroid]
               (v- (v* (inc alpha) centroid)
                   (v* alpha (peek simplex))))

             ;; Returns the point generated by reflecting the worst point across
             ;; the centroid, and then stretching it in that direction by a factor
             ;; of beta.
             (reflect-expand [simplex centroid]
               (v- (v* (inc (* alpha beta)) centroid)
                   (v* (* alpha beta) (peek simplex))))

             ;; Returns the point generated by reflecting the worst point, then
             ;; shrinking it toward the centroid by a factor of gamma.
             (reflect-contract [simplex centroid]
               (v- (v* (inc (* gamma alpha)) centroid)
                   (v* (* gamma alpha) (peek simplex))))

             ;; Returns the point generated by shrinking the current worst point
             ;; toward the centroid by a factor of gamma.
             (contract [simplex centroid]
               (v+ (v* (- 1 gamma) centroid)
                   (v* gamma (peek simplex))))

             ;; Returns a simplex generated by scaling each point toward the best
             ;; point by the shrink factor $\sigma$; ie, by replacing all
             ;; points (except the best point $s_1$) with $s_i = s_1 + \sigma (\s_i
             ;; - s_1)$.
             (shrink [[s0 & rest]]
               (let [scale-toward-s0 #(v+ s0 (v* sigma (v- % s0)))
                     s (into [s0] (map scale-toward-s0 rest))]
                 (sort-by-f s (mapv f s) dimension)))]

       (fn [simplex [f-best :as f-simplex]]
         (let [swap-worst (fn [elem f-elem]
                            (let [s  (conj (pop simplex) elem)
                                  fs (conj (pop f-simplex) f-elem)]
                              (sort-by-f s fs dimension)))
               f-worst    (peek f-simplex)
               f-butworst (peek (pop f-simplex))
               centroid   (centroid-pt simplex)
               reflected  (reflect simplex centroid)
               fr         (f reflected)]
           (cond
             ;; If the reflected point is the best (minimal) point so far, replace
             ;; the worst point with either an expansion of the simplex around that
             ;; point, or the reflected point itself.
             ;;
             ;; f(reflected worst) < f(best)
             (< fr f-best)
             (let [expanded (reflect-expand simplex centroid)
                   fe (f expanded)]
               (if (< fe fr)
                 (swap-worst expanded fe)
                 (swap-worst reflected fr)))

             ;; f(best) <= f(reflected worst) < f(second worst)
             ;;
             ;; Else, if the reflected worst point is better than the second worst
             ;; point, swap it for the worst point.
             (< fr f-butworst)
             (swap-worst reflected fr)

             ;; f(butworst) <= f(reflected worst) < f(worst)
             ;;
             ;; If the reflected point is still better than the worst point,
             ;; generated a point by shrinking the reflected point toward the
             ;; centroid. If this is better than (or equivalent to) the reflected
             ;; point, replace it. Else, shrink the whole simplex.
             (< fr f-worst)
             (let [r-contracted (reflect-contract simplex centroid)
                   frc (f r-contracted)]
               (if (<= frc fr)
                 (swap-worst r-contracted frc)
                 (shrink-simplex simplex)))

             ;; f(worst) <= f(reflected worst)
             ;;
             ;; Else, attempt to contrast the existing worst point toward the
             ;; centroid. If that improves performance, swap the new point; else,
             ;; shrink the whole simplex.
             :else
             (let [contracted (contract simplex centroid)
                   fc (f contracted)]
               (if (< fc f-worst)
                 (swap-worst contracted fc)
                 (shrink-simplex simplex))))))))))
```

And then this brings it all together:

```clojure
(defn multidimensional-minimize
  "Find the minimum of the function f: R^n -> R, given an initial point q ∈ R^n.

  Supports the following optional keyword arguments:

  :callback if supplied, the supplied fn will be invoked with the intermediate
  points of evaluation.

  :info if true, wraps the result with evaluation information.

  :adaptive? if true, the Nelder-Mead parameters for contraction, expansion,
  reflection and shrinking will be set adaptively, as functions of the number of
  dimensions. If false they stay constant.

  :alpha sets the reflection coefficient used for each step of Nelder-Mead.

  :beta sets the expansion coefficient used for each step of Nelder-Mead.

  :gamma sets the contraction coefficient used for each step of Nelder-Mead.

  :sigma sets the shrink coefficient used for each step of Nelder-Mead.

  :maxiter Maximum number of iterations allowed for the minimizer. Defaults to
  200*dimension.

  :maxfun Maximum number of times the function can be evaluated before exiting.
  Defaults to 200*dimension.

  :simplex-tolerance When the absolute value of the max difference between the
  best point and any point in the simplex falls below this tolerance, the
  minimizer stops. Defaults to 1e-4.

  :fn-tolerance When the absolute value of the max difference between the best
  point's function value and the fn value of any point in the simplex falls
  below this tolerance, the minimizer stops. Defaults to 1e-4.

  :zero-delta controls the value to which 0 entries in the initial vector are
  set during initial simplex generation. Defaults to 0.00025.

  :nonzero-delta factor by which entries in the initial vector are perturbed to
  generate the initial simplex. Defaults to 0.05.

  See Gao, F. and Han, L.
      Implementing the Nelder-Mead simplex algorithm with adaptive
      parameters. 2012. Computational Optimization and Applications.
      51:1, pp. 259-277

  I gratefully acknowledge the [Python implementation in
  SciPy](https://github.com/scipy/scipy/blob/589c9afe41774ee96ec121f1867361146add8276/scipy/optimize/optimize.py#L556:5)
  which I have imitated here.
  "
  [func x0 & {:keys [callback info]
              :or {callback (constantly nil)}
              :as opts}]
  (let [dimension     (count x0)
        [f-counter f] (counted func)
        step          (step-fn f dimension opts)
        stop?         (stop-fn f-counter dimension opts)
        simplex       (initial-simplex x0 opts)
        f-simplex     (mapv f simplex)]
    (loop [[[s0 :as simplex] [f0 :as f-simplex]] (sort-by-f simplex f-simplex dimension)
           iteration 0]
      (callback s0)
      (if (stop? iteration simplex f-simplex)
        (if info
          {:result s0
           :value f0
           :iterations iteration
           :fncalls @f-counter}
          s0)
        (recur (step simplex f-simplex)
               (inc iteration))))))
```

sritchie (commented):
Amazing

sritchie:
Okay @littleredcomputer , sorry again for hijacking - I've just pushed my changes, so I'll kick this back to you.

Here's the diff: https://github.com/littleredcomputer/sicmutils/pull/102/commits/325b128b58aff66738373dccef9558cc384b8da7

I've:

- changed to the style in the gist
- split `stop-fn` into a `convergence?` check and a separate stop check, so that `:convergence?` gets returned in the results gets returned in the results
- added a `:pre` and `:post` condition that `f-simplex` stays sorted
- slightly modified the tests to use this `same` library we're using for testing, which lets us write:

```clojure
(with-comparator near
        (is (ish? [0.17285378 -0.524869316 0.487525860]
                  v)))
```

without unpacking the list elements ourselves.

- added back the ability to pass `:info` and get a more detailed response.

I think this is looking great, but take a look and make any changes you like before approving and merging.

What do you recommend I take on? I think we still need `definite-integral` support to get a few of the earlier examples to run. Single-dimensional minimize is good too.

If you take the `odex` integration, then we should be... there! amazing.

littleredcomputer:
I know, it's so close, and I want to do all this because I have not been
pulling my weight.

The brent 1-dimensional minimize in SciPy seems annoyingly elaborate, but
everyone else uses it, so we might give it the same treatment.

For definite-integral, my plan was to just run odex in one-dimensional mode
and maybe kill two birds with one stone.

Thank you for the thorough work on multidimensional!

littleredcomputer:
actually using the ODE solver for a definite integral solver? Never done
it, not entirely sure it's a good idea or even possible. Will have to think
about that one.

Another possibility would be to lift the implementation from the original
scmutils for definite integral. Haven't dug into that stuff in a while :)

PR #103, opened 2020-09-09
Various minimization routines ported from Scipy (in progress)

sritchie:
This PR implements in native Clojure(script):

- Scipy's `bracket` procedure, for generating an interval that contains a function's minimum
- the golden section univariate minimizer
- Brent's method

These were a heavy lift, mostly because Brent's method has been ported over and over from Fortran and was difficult to decode. The wikipedia page was quite misleading, since it's for Brent's root-finding method... https://en.wikipedia.org/wiki/Brent%27s_method

AND it's wrong! I tried to heavily document this implementation of Brent's method to help anyone else that comes along.


src/sicmutils/numerical/unimin.cljc
```diff
@@ -0,0 +1,450 @@
...
+(defn- ascending-by
+  "Returns the points ordered as f(a) < f(b)"
+  [f a b]
+  (let [fa (f a) fb (f b)]
+    (if (< fa fb)
+      [[a fa] [b fb]]
+      [[b fb] [a fa]])))
+
+(defn- lagrange-interpolating-polynomial
+  "Generates a lagrange interpolating polynomial that fits all of the supplied
```
littleredcomputer, line 41: I'm warming up to the idea of keeping the polynomial library general. It was a while to get to that point since so much of the performance battle had to be fought there and as we see from the ch7 test there's more to learn.

src/sicmutils/numerical/unimin.cljc
```diff
@@ -0,0 +1,450 @@
...
+  ([f] (bracket-min f {}))
+  ([f {:keys [xa xb max-tries]
+       :or {xa 0.0
+            xb 1.0
+            max-tries 1000}
+       :as opts}]
+   (let [[f-counter f] (u/counted f)
+         step (bracket-step-fn f opts)
+         stop-fn (fn [[xa fa :as a] [xb fb :as b] [xc fc :as c] iteration]
+                   ;; TODO - properly distinguish convergence from the other
```
littleredcomputer, line 157: I had started to do Brent from spicy and thought about this too. My conclusion at the time was "throw here, since if the bracket-min exhausts its budget you must really have crossed over into Lulu land, and the outer algorithm is not likely to be able to get this back on track," while leaving the outer (Brent) iteration count in place in the usual way. One could also imagine sharing Brent's iteration budget with this lower routine. In the end, I think it would be very unlikely to see a failure here since IIUC the interval width grows like exp(iteration)

littleredcomputer (commented):
beautiful

sritchie:
@littleredcomputer I need to reorganize and push but I went down a serious rabbit hole with Brent’s, all the way back to the original book... and I’ve learned that the Wikipedia algorithm is wrong! It’s all much simpler than the SciPy code makes it seem, too. I’ll get this organized and pushed, and written up, in the next day or so. I’m excited to have a few nice examples here of optimizers so we can start to shake out a common interface...

sritchie:
@littleredcomputer I need to add more tests and split this into three... but take a look, I bet you'll like this: https://github.com/littleredcomputer/sicmutils/blob/32332c63f0e25b53f3c7e4f99e2eadff1d068e92/src/sicmutils/numerical/unimin/brent.cljc

src/sicmutils/numerical/unimin.cljc
sritchie: Agree here. I've finally got these all polished and put up, so please take a look when you get a chance and let me know what you think!

The final steps are `definite-integral` and ODE support...

sritchie:
Closing, since we now have #104, #105, #106, #107 and #108.

PR #104, opened 2020-09-16
Port scipy's auto-bracketing + scmutils version

sritchie:
This PR:

- ports `scipy.optimize.bracket` over to `bracket-min` and `bracket-max`: https://github.com/scipy/scipy/blob/v1.5.2/scipy/optimize/optimize.py#L2450-L2568
- ports the simpler bracketing logic from `scmutils`
- adds tests for each of the above

This is a component we can use to automatically generate brackets for the minimizers I'll bring in later.

EDIT: Looks like the tests are failing because I moved a fn to `golden.cljc`, which I introduce only in #105. I'm going to beg for a review here where I don't untangle this, because I introduce a dependency BACK to bracket in the golden tests... so please review, but verify that the tests pass in #105.


src/sicmutils/numerical/unimin/bracket.cljc
```diff
@@ -0,0 +1,278 @@
...
+        q (* 2.0 v)]
+    (if (pos? q)
+      [(g/negate p) q]
+      [p (g/abs q)])))
+
+(defn parabolic-step
+  "Fits a parabola through all three points, and returns the coordinate of the
+  minimum of the parabola.
+
+  If the supplied points are colinear, returns a point that takes a large jump
```
littleredcomputer, line 85: sp: collinear

littleredcomputer (approved):
Your care with this is inspiring

PR #106, opened 2020-09-16
Implement Brent's method for fn minimization in native clj

sritchie:
This was a heavy lift, mostly because Brent's method has been ported over and over from Fortran and was difficult to decode. The wikipedia page was quite misleading, since it's for Brent's root-finding method... https://en.wikipedia.org/wiki/Brent%27s_method

AND it's wrong! I tried to heavily document this implementation of Brent's method to help anyone else that comes along. I ported a bunch of tests from `BrentOptimizer`, added more `test.check` tests and compared outcomes with the `scipy` implementation.


src/sicmutils/numerical/unimin/brent.cljc
```diff
@@ -0,0 +1,359 @@
...
+  "For convenience, we also provide the sister-procedure for finding the maximum
+  of a unimodal function using Brent's method.
+
+  Negate the function, minimize, negate the result."
+  [f a b opts]
+  (let [-f (comp g/negate f)]
+    (-> (brent-min -f a b opts)
+        (update-in [:value] g/negate))))
+
+#?(:clj
```
littleredcomputer, line 298: any objection to removing commons as a dependency??

test/sicmutils/numerical/unimin/brent_test.cljc
```diff
@@ -0,0 +1,189 @@
...
+          ;; The function has local minima at -0.27195613 and 0.82221643.
+          (is (zeroish? ((D f) max1)) "verify max1 is a local maximum.")
+
+          (is (ish? {:result max1
+                     :value (f max1)
+                     :converged? true
+                     :iterations 8
+                     :fncalls 9}
+                    (b/brent-max f 0.2 0.3 {:maxeval 200}))
+              "local max for the quintic."))))))
```
littleredcomputer, line 189: I suppose you could generate N random reals c_i in the range [-10, 10], form the polynomial ∏(x-c_i), and see if the minimum is close to one of the chosen roots. No doubt there are edge cases that might force some conditioning on the constants (not too close together perhaps)

littleredcomputer (approved):

src/sicmutils/numerical/unimin/brent.cljc
sritchie: Nope, except that we still need it for the Complex implementation, and of course the methods we haven't ported yet. I'm going to take on `definite-integral` next. Any progress with the odex implementation?

PR #109, opened 2020-09-16
bigint compatiblity with floating point

sritchie:
This PR extends `BigInt` arithmetic in JS to be compatible with floating point numbers. We probably want to be promoting floats to BigDecimal in this case... but this is fine for now.


test/sicmutils/numerical/unimin/bracket_test.cljc
```diff
@@ -25,6 +25,7 @@
             [same :refer [zeroish? ish? with-comparator]
              #?@(:cljs [:include-macros true])]
             [sicmutils.calculus.derivative :refer [D]]
+            [sicmutils.function]
```
sritchie, line 28: this was necessary to register the multimethods that get `D` working.

littleredcomputer (approved):

PR #111, opened 2020-10-02
Add suite of tools for definite integration [in progress]

sritchie:
Lots here that I'll need to redo and split up!


src/sicmutils/numerical/interpolate/richardson.cljc
```diff
@@ -0,0 +1,217 @@
...
+  ;; That's from numerical recipes. again, the KEY POINT is that if we want to
+  ;; use this method, we need to do the squared trick for our trapezoidal
+  ;; shenaningans, and for our midpoint method.
+  ;;
+  ;; http://phys.uri.edu/nigh/NumRec/bookfpdf/f4-3.pdf
+  ;;
+  ;; If we did NOT have those - if we used these tricks for the left and right
+  ;; methods - we'd have to extrapolate on `h` itself, on the actual window
+  ;; width, not the squared window width. No big deal.
+  )
```
littleredcomputer, line 217: This has been fascinating to read. It might make a cool notebook, eh?

src/sicmutils/numerical/quadrature.cljc
```diff
@@ -0,0 +1,152 @@
...
+;;; Default compile integrand
+(def ^:dynamic *compile-integrand?* true)
+
+;;; Default error specification
+(def ^:dynamic *definite-integral-allowable-error* 1.0e-11)
+
+;;; TODO what the hell is this? Default ditherer off (see rational.scm)
+(def ^:dynamic *quadrature-neighborhood-width* false)
+
+(def methods
```
littleredcomputer, line 47: The SICP style of UI for things like this cuts aggressively against the Clojure grain, I think. WDYT? On the other hand, when I learned it from the book way back when, I thought it was a gas, and you can do cool things with it.

src/sicmutils/numerical/quadrature/infinite.cljc
```diff
@@ -0,0 +1,155 @@
...
+
+(defn inv-square-upper
+  "Change of variables for singularity on the upper side."
+  [method gamma]
+  (fn [f a b opts]
+    (let [f' (fn [t] (* t (f (- b (* t t)))))]
+      (* 2 (method f' 0 (g/sqrt (- b a)) opts)))))
+
+(defn evaluate-infinite-integral
+  "Handles all integrals with an infinite endpoint. I think there are a lot of
```
littleredcomputer, line 93: Maybe I need a new edition of Press. I have 2ed on dead tree. The C++ style found there is maddening.

src/sicmutils/numerical/quadrature/rational.cljc
```diff
@@ -0,0 +1,200 @@
...
+              (- b a))))
+    (+ a (* 0.5 (- b )))))
+
+;; These all adaptively drop down if they're not able to converge a rational fit
+;; within 10 steps.
+;;
+;; They recommend 8 in that text.
+(def ^:dynamic *integrate-n* 10)
+
+(defn integrate-closed-closed [f a b eps]
```
littleredcomputer, line 178: This is awesome, I hadn't thought of merging an integrator designed to work with an open endpoint with bisection this way.

littleredcomputer (approved):
It's like a survey paper, I'm very impressed

src/sicmutils/numerical/elliptic.cljc
```diff
@@ -0,0 +1,80 @@
...
+                     e2)
+                  (* c4 e3))
+               (g/sqrt ave))))))))
+
+(defn elliptic-f [phi k]
+  "Legendre elliptic integral of the first kind F(φ, k).
+   See W.H. Press, Numerical Recipes in C++, 2ed. eq. 6.11.19
+
+  Page 260 here: http://phys.uri.edu/nigh/NumRec/bookfpdf/f6-11.pdf"
+  (let [s (g/sin phi)]
```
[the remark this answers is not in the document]
sritchie: You are totally right (and sorry for my delay here, I've finally got the abstractions "figured out" and will start shuffling this stuff into PRs! I also dropped the ball on our meeting due to some personal travel, so apologies there too.

It is all much faster when we skip the generics and go straight to native operations. I did this originally so I could pass in symbols, and then show that the simpler and more obfuscated methods were equivalent for a few points (over in the integral stuff, not here...)

I wonder if, instead of doing that, we could use `with-redefs` to override `+, -, /, *` temporarily in numerical methods and force them generic, just for tests.

That got me thinking - can we do it the other way? Could we write a macro that looks like the values of the multimethods just once, and then rebinds them all (to prevent lookups) inside the macro body, if you can guarantee you're only going to use a specific type?

src/sicmutils/numerical/quadrature.cljc
sritchie: yeah, I think you're probably right. The integrator with mutable methods is pretty weird, and it's not clear why you wouldn't just call the constructor again. I think it's still nice to at least provide the `refman` interface, even if we go a little more wild under the hood...

src/sicmutils/numerical/quadrature/infinite.cljc
sritchie: It is pretty good, but I've definitely found that I have to be careful porting the code over directly. We'll add Press to our list of books to port :)

src/sicmutils/numerical/quadrature/rational.cljc
sritchie: Thanks to Sussman we've got it! One strange thing is that for `open-closed` and `closed-open` they try full open first, then bisect out to `open-closed` and `closed closed`. Any reason you can think of that, for those two, we should just bisect immediately, so that we force the proper treatment of the closed endpoint?

src/sicmutils/numerical/derivative.cljc
```diff
@@ -0,0 +1,78 @@
+;;
+;; Copyright © 2017 Colin Smith.
```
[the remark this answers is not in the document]
sritchie: We can redo the copyrights when we move this (love your suggestion) to a separate numerical package! Any ideas for a github organization name we could create to house this stuff? `sicmutils` doesn't quite feel right since we're moving beyond utilities for those books...

sritchie:
Phew... closing in favor of:

#118, #119, #120, #121, #123, #125, #126, #127, #128, #129.

PR #112, opened 2020-10-06
Stream utilities and file moves (definite integral #1)

sritchie:
This PR:

- moves `nelder-mead` into its own namespace
- fixes a few tests that had been broken, whoops!
- moves the elliptic functions to their own namespace
- adds a few stream utilities that I need for later, in definite integral land
- converts the elliptic stuff back to native cos, sin etc methods for better performance
- fixes a bug where we were only passing the value to the callback in nelder mead. previously (and in tests we passed iteration, value AND `fx`.

One nice thing here is some routines for summing sequences of floating point numbers and trying to minimize error a bit, using compensated summation:

https://en.wikipedia.org/wiki/Kahan_summation_algorithm


sritchie:
Hey @littleredcomputer , I'm going to go ahead and merge these so I don't get too far ahead, but I'd love your comments! My goal here is to see if we turn these into sort of... networked literate programming documents with embedded visualizations, that ALSO are the actual code for the library. That should explain the heavy doses of comments.

littleredcomputer:
Thank you. I looked at it and will give it more formal treatment.

I purchased Hairer & Wanner's book from Springerlink, after figuring out
that they discuss GBS and ODEX.F in depth, complete with discussions of the
step size iterations and adaptivity, with graphs of instrumented solvers
achieving convergence to demonstrate the effectiveness of the method.

This has made me aware that, rather than working in the Fortran style, your
work could indeed be the basis of an implementation which would be just as
good and rather more functional, if not completely so.

I still want to have a pure clj reference implementation that is bug-alike
to odex.f, but the endpoint now for me is to site it atop your Richardson
extrapolation code once I understand it all well enough.

I love functional programming but things like counting function invocations
with an atom or throwing an exception as a way of escaping from a deep
entanglement one has come to regret don't bother me too much.

sritchie:
@littleredcomputer totally agree, I think that is the beauty of the mixed approach....  being able to use an atom for function counts etc.

The latest "bulirsch-stoer" integrator should be very similar to what you'll need for functional odex: #126 

it has the nice advantage of being able to take advantage of the trapezoid and midpoint method's reuse of function evaluations. I wonder if there's anything here to steal for the more hardcore, stateful implementation.

I think it would be extraordinary to have a composable, understandable, functional implementation to compare the more hardcore version to.

Also, I was thinking... you know how the current stuff has to do a bunch of copies from:

- derivative evaluations into the output buffer
- back OUT from the output buffer, if you want to read the results and pass them out?

Do all of those copies mean that, for something like an animator, where we are looking at tons of intermediate results, that the functional version might actually be faster?

Or, maybe internally we should just return the array full-stop and create a new one every time the "observe" function is called.

PR #113, opened 2020-10-07
Polynomial interpolation + tests (definite integral #2)

sritchie:
This PR adds (the basis for) an interactive, computational essay on polynomial interpolation using:

- lagrange polynomial interpolation, and
- two versions of Neville's algorithm

I've also implemented each in two different styles, and abstracted them into a form that will extend nicely to richardson interpolation and rational function interpolation.

I'll go less turbo on those two PRs, and refer folks back to the polynomial namespace.


src/sicmutils/util/stream.cljc
```diff
@@ -65,12 +65,12 @@
   `identity`.
     "
   [f & {:keys [present init]
-        :or {present identity
-             init (f)}}]
```
sritchie, line 69: annoyingly, this evaluates `(f)` EVEN if you supply `init`!

src/sicmutils/numerical/unimin/bracket.cljc
```diff
@@ -32,28 +32,6 @@
       [[a fa] [b fb]]
       [[b fb] [a fa]])))
 
-(defn lagrange-interpolating-polynomial
-  "Generates a lagrange interpolating polynomial that fits all of the supplied
```
sritchie, line 36: I moved this to the new poly namesapce.

src/sicmutils/util/stream.cljc
littleredcomputer: Sometimes. Depends on what you name the keywords. While porting ODEX.F, which has many options, I pulled my hair out trying to figure out why providing an init value that used the value of some other option worked for one option but not another. Macroexpand revealed that (apparently) all the options are put in a bag by the compiler and then a big `let*` is emitted... which would work fine if the order of the keywords were preserved.

littleredcomputer: Oh, upon second reading I see that you're making a different complaint. Yeah, it's too bad we don't have the semantics of the `(or)` form here. We have found that the `:keys:or` thing is not exactly industrial strength :)

PR #115, opened 2020-10-07
Add richardson extrapolation, special case of poynomial extrapolation (defint #4)

sritchie:
This PR adds an implementation of Richardson extrapolation - we need this for the Romberg integration method, specifically.

I built this one up as a "computational essay" too. I'm looking forward to finding some platform where we can share these lessons without burying them in a source file!


src/sicmutils/numerical/interpolate/richardson.cljc
```diff
@@ -0,0 +1,244 @@
...
+         xs
+         (rest xs))))
+
+;; If we start with the original sequence, we can implement Richardson
+;; extrapolation by using Clojure's `iterate` with the `accelerate-sequence`
+;; function to generate successive columns in the "Richardson Tableau". (This is
+;; starting to sound familiar to the scheme for polynomial interpolation, isn't
+;; it?)
+;;
+;; To keep things general, let's take a general sequence `ps`, defaulting to the
```
littleredcomputer, line 176: very cool. You have to hand it to GJS for hitting on the arithmetic progression as a simplifying vehicle

littleredcomputer (approved):

PR #117, opened 2020-10-09
Riemann quadrature, ie, inefficient (but informative!) definite integrals (definite integral #1)

sritchie:
This PR (still pending) implements various quadrature methods in a simple way that's easy to read and understand. My plan is to keep these here, and then develop them into more efficient versions so folks can follow what's going on.

This is more meat extracted from #111 .


sritchie:
Closing, as this is now superceded by #118, #119, #120 , #121.

PR #118, opened 2020-10-13
Custom column selection for tableau methods, fix for Richardson (integrals #1)

sritchie:
This PR:

- adds the ability to select specific columns from the tableau methods. This is important for recreating Simpson's method, etc, from the more general methods.
- fixes a bug where Richardson extrapolation would returns an endless sequence of `nil` if its input seq ran out.

brace yourself for more integral PRs!

### DIFF
--- a/src/sicmutils/numerical/interpolate/rational.cljc
+++ b/src/sicmutils/numerical/interpolate/rational.cljc
@@ -114,12 +114,31 @@
   The implementation follows Equation 3.2.3 on on page 105 of Press:
   http://phys.uri.edu/nigh/NumRec/bookfpdf/f3-2.pdf.
 
+  ## Column
+
+  If you supply an integer for the third (optional) `column` argument,
+  `bulirsch-stoer` will return that /column/ offset the interpolation tableau
+  instead of the first row. This will give you a sequence of nth-order
+  polynomial approximations taken between point `i` and the next `n` points.
+
+  As a reminder, this is the shape of the tableau:
+
+   p0 p01 p012 p0123 p01234
+   p1 p12 p123 p1234 .
+   p2 p23 p234 .     .
+   p3 p34 .    .     .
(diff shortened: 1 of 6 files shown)


src/sicmutils/numerical/interpolate/rational.cljc
```diff
@@ -114,12 +114,31 @@
...
 
+  ## Column
+
+  If you supply an integer for the third (optional) `column` argument,
+  `bulirsch-stoer` will return that /column/ offset the interpolation tableau
+  instead of the first row. This will give you a sequence of nth-order
+  polynomial approximations taken between point `i` and the next `n` points.
+
+  As a reminder, this is the shape of the tableau:
+
```
littleredcomputer, line 125: How are these indices used? Is there a get-in type of thing for the tableau where these small integer indices could be specified... i.e., what does this look like as a seq?

littleredcomputer (approved):

src/sicmutils/numerical/interpolate/rational.cljc
sritchie: The tableau ends up looking like a lazy sequence of lazy sequences (the columns), and we have:

- the ability to pull a particular column, and
- the ability to pull the top row.

(If you wanted NOT the top row, you'd drop the first point from the input sequence.)

Taking the first row is like doing Romberg integration - your polynomials become higher and higher in order.

As you can see from the Simpson etc PR, the columns here are ACTUALLY like fitting higher-order polynomials inside the intervals from the original sequence!

So if you do a polynomial tableau and pass in a sequence trapezoid estimates, the second column gives you simpson's rule. (Fit a like between 01, 12, 23, etc... but the estimates were already lines, so this gives you a quadratic.)

Does that make sense? I think my function signature is bad here, but the ability to pull

- columns or
- the first row

is what's important.

PR #119, opened 2020-10-13
{left, right, upper, lower} Riemann integrators, essay + tests (integrals #2)

sritchie:
This PR adds efficient implementations of left, right, upper and lower Riemann sums. Alone these aren't "suitable" for production use... but with Richardson extrapolation they're great.

In fact, the method that claims to be the midpoint method in Apache Commons is actually just a Left Riemann sum, without the left-most evaluation point! Or a right without the right point, equivalently. A major bug.

### DIFF
--- a/src/sicmutils/numerical/derivative.cljc
+++ b/src/sicmutils/numerical/derivative.cljc
@@ -1,0 +1,577 @@
+;;
+;; Copyright © 2020 Sam Ritchie.
+;; This work is based on the Scmutils system of MIT/GNU Scheme:
+;; Copyright © 2002 Massachusetts Institute of Technology
+;;
+;; This is free software;  you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3 of the License, or (at
+;; your option) any later version.
+;;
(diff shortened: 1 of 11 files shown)


sritchie:
@littleredcomputer the pieces are coming together :) This ended up being a bigger project than I'd thought. But now we have some pretty nice plug-in components that should come in handy with ODEX.

Hopefully these ultra-documented namespaces aren't too obnoxious. I wanted to generate some fodder to put into a system like Maria. Let me know what you think!

littleredcomputer (approved):
This was fascinating. As I read it, I wondered: is it really worth the effort to reuse function evaluations when the partition series is not 2^k?

sritchie:
@littleredcomputer I think it is, since it comes along for free - I did that specifically because, in the bulirsch-stoer method, we use these funky step-size jumps where it's not so obvious how to explicitly save function evaluations.
```
(def bulirsch-stoer-steps
  (interleave
   (us/powers 2 2)
   (us/powers 2 3)))
```

But if the underlying trapezoid or midpoint methods can just... do it automatically... then we get a nice boost. I added a test somewhere in here that shows that we save 30% of evaluations or so in this case.

PR #120, opened 2020-10-13
incremental and accelerated Midpoint integrators, essay + tests (integrals #3)

sritchie:
This PR implements the midpoint method, building on functions we developed exploring Riemann sums and Richardson extrapolation.

I was able to write an efficient implementation that reuses former evaluations as it ups the number of grid points. Press mentions it in Numerical Recipes, but I couldn't find it anywhere else!

As I mentioned in #119 ... the apache commons method is actually "not" the midpoint method. You can see in the code that they refine by doubling the number of interior points: http://commons.apache.org/proper/commons-math/javadocs/api-3.5/src-html/org/apache/commons/math3/analysis/integration/MidPointIntegrator.html#line.37

This doesn't make sense, as you'll see when you read the namespace. What's really going on is they're taking a left (or right) Riemann sum without any of the endpoints. It's probably not that far off... but if you don't think that those methods are valuable in practice, then don't use the Apache Commons implementation.

### DIFF
--- a/src/sicmutils/numerical/derivative.cljc
+++ b/src/sicmutils/numerical/derivative.cljc
@@ -1,0 +1,577 @@
+;;
+;; Copyright © 2020 Sam Ritchie.
+;; This work is based on the Scmutils system of MIT/GNU Scheme:
+;; Copyright © 2002 Massachusetts Institute of Technology
+;;
+;; This is free software;  you can redistribute it and/or modify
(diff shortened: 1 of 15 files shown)


littleredcomputer (approved):
This is really cool.

sritchie:
Thanks @littleredcomputer ! I think you'll like the more advanced ones, built out of these tighter components.

PR #121, opened 2020-10-13
incremental and accelerated Trapezoid integrator, essay + tests

sritchie:
This PR implements the trapezoid method (with an incremental version as default, and an optional to accelerate), plus an essay describing how it was all done.

If you pass `:accelerate? true`, you've gained Romberg integration without any special effort.


src/sicmutils/numerical/quadrature/trapezoid.cljc
```diff
@@ -0,0 +1,301 @@
...
+
+#_
+(- Math/PI (pi-estimator* 10))
+;; => 0.0016666646826344333
+
+;; 10000 slices gets us closer:
+
+#_
+(< (- Math/PI (pi-estimator* 10000))
+   1e-8)
```
littleredcomputer, line 127: presumably the result is `true`

src/sicmutils/numerical/quadrature/trapezoid.cljc
```diff
@@ -0,0 +1,301 @@
...
+
+  If `n` is a sequence, the resulting sequence will hold an estimate for each
+  integer number of slices in that sequence."
+  ([f a b] (trapezoid-sequence f a b 1))
+  ([f a b n]
+   (let [S      (trapezoid-sum f a b)
+         next-S (qr/Sn->S2n f a b)]
+     (qr/incrementalize S next-S 2 n))))
+
+;; The following example shows that for the sequence $2, 3, 4, 6, ...$ (used in
```
littleredcomputer, line 225: As a reader, if I weren't following along, I would wonder why you didn't make the comparison between the incremental version and the non-incremental version in the straight-up 2x case just as a check on the correctness of Sn->S2n here. Then you could show that the B/S sequence lies properly between the two cases in evaluation count, since the other two cases form the maximum and minimum reuse cases

littleredcomputer (approved):

src/sicmutils/numerical/quadrature/trapezoid.cljc
sritchie: yes, I didn't know how much to fake these little process outputs... I figure that when I get this going in Maria etc I can have the code's result just be there.

src/sicmutils/numerical/quadrature/trapezoid.cljc
sritchie: Good call! I added this block: https://github.com/littleredcomputer/sicmutils/pull/121/commits/9e98f26152f0417368f34d7f14a1035b9e7e772c

PR #123, opened 2020-10-14
Add integrators that pull from the Richardson Tableau (integrals #5)

sritchie:
This PR adds:

- Milne's Method (midpoint method, one richardson step)
- Simpson's Method (trapezoid, one richardson step)
- Simpson's 3/8 Method (trapezoid, one richardson step with 3x slices per step)
- Boole's Method (trapezoid, TWO richardson steps!)
- Romberg Integration (open and closed, ie, midpoint and trapezoid methods, first row of the Richardson tableau)

Next comes the code to make any of these adaptive!


littleredcomputer (approved):
Seeing that Romberg has an open and a closed form makes me wonder if we want to use the bisection technique for the half-open case. I will be curious to see how the open form handles divergence on one side: is it awesome, or does it approach the infinite endpoint too slowly to gather the needed area contributions? One thing about the bisection is it allows you to grind the problematic endpoint harder, without doing a lot of busy work in the smooth sections.

sritchie:
Totally agree that the adaptive version of all of these is going to be better, @littleredcomputer . I don't have a stable of test functions in my brain that I know to throw against these, so I'd love some help now that the interface is nice and usable!

I've made an issue here to track this: https://github.com/littleredcomputer/sicmutils/issues/136 Let me know what you think.

PR #125, opened 2020-10-20
put all integrators behind a common `defintegrator` interface (integrals #6)

sritchie:
This PR adds a `defintegrator` macro that makes it easier to define abstracted integrators with nice docstrings. I'll add a few comments inline about other small changes I made to tidy up the interface. Getting there!


src/sicmutils/numerical/quadrature/boole.cljc
```diff
@@ -18,31 +18,71 @@
...
-            [sicmutils.numerical.interpolate.richardson :as ir]
-            [sicmutils.util.stream :as us]))
+  (:require [sicmutils.numerical.quadrature.common :as qc
+             #?@(:cljs [:include-macros true])]
+            [sicmutils.numerical.quadrature.trapezoid :as qt]
+            [sicmutils.numerical.interpolate.richardson :as ir]))
 
 ;; ## Boole's Rule
 ;;
-;; Commonly mis-spelled as "Bode's Rule"!
```
sritchie, line 27: whoops, I'd forgotten docs here!

src/sicmutils/numerical/quadrature/common.cljc
```diff
@@ -0,0 +1,198 @@
...
+;;   (integrate f a b))
+
+(def ^:dynamic *roundoff-cutoff* 1e-14)
+
+;; ## Intervals
+;;
+;; Implementations of the various intervals used by the adaptive integral
+;; interface. By default, integration endpoints are considered /open/.
+
+(def open        [::open ::open])
```
sritchie, line 54: these come into play later, when we implement the adaptive wrapper.

src/sicmutils/numerical/quadrature/common.cljc
```diff
@@ -0,0 +1,198 @@
...
+;; namespaces can use to create an "integrator" from:
+;;
+;; - a fn that can estimate the area of a single integration slice, and
+;; - a fn that can return a sequence of progressively finer estimates.
+;;
+;; The first function is called in the case that the integration range $(a,
+;; b)$ (open or closed) is too fine for subdivision. The second function takes
+;; over in all other (most!) cases.
+
+(defn- narrow-slice?
```
sritchie, line 115: This also only becomes REALLY revelant when we get adaptive going. But it is a nice feature to share with all integrators now!

src/sicmutils/numerical/quadrature/common.cljc
```diff
@@ -0,0 +1,198 @@
...
+  False otherwise. This inequality measures how close the two floating point
+  values are, scaled by the sum of their magnitudes."
+  [a b cutoff]
+  (let [sum (+ (Math/abs a)
+               (Math/abs b))]
+    (or (<= sum cutoff)
+        (<= (Math/abs (- b a))
+            (* cutoff sum)))))
+
+(defn make-integrator-fn
```
sritchie, line 130: users can call this directly if they want to generate an integrator without using the `defintegrator` macro.

src/sicmutils/numerical/quadrature/midpoint.cljc
```diff
@@ -150,16 +152,25 @@
...
+  `:accelerate?`: if supplied (and `n` is a number), attempts to accelerate
+  convergence using Richardson extrapolation. If `n` is a sequence this option
+  is ignored."
+  ([f a b] (midpoint-sequence f a b {:n 1}))
+  ([f a b {:keys [n accelerate?] :or {n 1}}]
    (let [S      (qr/midpoint-sum f a b)
-         next-S (Sn->S3n f a b)]
-     (qr/incrementalize S next-S 3 n))))
+         next-S (Sn->S3n f a b)
+         xs     (qr/incrementalize S next-S 3 n)]
```
sritchie, line 170: I pushed the acceleration options etc back into the `sequence` functions themselves; this makes them all nice and customizable, and also makes it easier to build a common interface around them all.

sritchie: All the sequence functions now also take a shared `opts` dictionary, same as the integrators. Curious to hear what you think, @littleredcomputer !

src/sicmutils/numerical/quadrature/riemann.cljc
```diff
@@ -485,43 +528,116 @@
...
          next-S (Sn->S2n f a b)]
-     (incrementalize S next-S 2 n))))
+     (-> (incrementalize S next-S 2 (:n opts 1))
+         (accelerate opts)))))
+
+;; `lower-sequence` and `upper-sequence` are similar. They can't take advantage
+;; of any incremental speedup, so we generate a sequence of `n`s internally and
+;; map `lower-sum` and `upper-sum` directly across these.
+
+(defn lower-sequence
```
sritchie, line 584: I made these their own sequence functions with comments, to make it simpler to call `defintegral` on them, and bring them up to par with the other integrators (even though no one should use these :) )

src/sicmutils/numerical/quadrature/romberg.cljc
```diff
@@ -96,14 +100,30 @@
...
+  If supplied, `:n` (default 1) specifies the initial number of slices to use."
+  ([f a b] (closed-sequence f a b {}))
+  ([f a b {:keys [n] :or {n 1} :as opts}]
    {:pre [(number? n)]}
-   (-> (qt/trapezoid-sequence f a b n)
+   (-> (qt/trapezoid-sequence f a b opts)
        (ir/richardson-sequence 2 2 2))))
 
-(defn open-integral
+(defn romberg-sequence
```
sritchie, line 112: I added a function here that dispatches to either open or closed depending on the contained interval, so we can exercise that `common` feature now.

littleredcomputer: Some of the underlying techniques have open variants and some do not, IIUC. Should the individual techniques advertise half- or full-open capability if they have it, and then map it in on the appropriate side? It could also be an error to request an open integration for an underlying method which does not know how to avoid the endpoints

src/sicmutils/numerical/quadrature/midpoint.cljc
littleredcomputer: This looks cool and I can see the direction where all this is heading. An opts dictionary is surely the way to go. I suppose it starts at the top, and is passed down to the lower levels, each level will look at different things... is that the model?

littleredcomputer (approved):

src/sicmutils/numerical/quadrature/common.cljc
```diff
@@ -0,0 +1,198 @@
...
+;; An integrator will estimate a single slice directly. Else, it will attempt to
+;; converge a sequence.
+;;
+;; NOTE - we don't have an interface yet to bind this dynamic variable. bind it
+;; manually to modify the cutoff for a specific call to some integrator:
+;;
+;; (binding [*roundoff-cutoff* 1e-6]
+;;   (integrate f a b))
+
+(def ^:dynamic *roundoff-cutoff* 1e-14)
```
littleredcomputer, line 47: Maybe this should go in the options dict?

src/sicmutils/numerical/quadrature/riemann.cljc
littleredcomputer: I like that you have done this

littleredcomputer (approved):

src/sicmutils/numerical/quadrature/midpoint.cljc
sritchie: That's correct!

src/sicmutils/numerical/quadrature/romberg.cljc
sritchie: I think this is a great idea. I've added #136 so we can track this. Might be a great place to get Spec involved.

src/sicmutils/numerical/quadrature/common.cljc
sritchie: I put it in there too, but left this one around, since I wasn't sure if I wanted to completely kill the ability to dynamically specify these options. Too much reliance on `scmutils`. What do you think? Should we kill these?

PR #126, opened 2020-10-20
"Bulirsch-Stoer" definite integral implementation (integrals #7)

sritchie:
This PR adds an implementation of the rational-function-interpolating integration that scmutils actually uses internally. They pair this with an adaptive wrapper that

- attempts to converge in 10 steps, and
- if convergence isn't reached, splits the interval and attempts again on each side.

I know I can write this adaptive logic in a more generic way, so that it will work with ANY integrator. To start, I'm going to add the open and closed versions of their "bulirsch-stoer" definite integrators.

They're called that because

- they use rational function interpolation instead of polynomial, like the richardson algos
- they use a non-geometrically-increasing step sequence, so they can't take the Richardson optimization.

This is what's going on inside of Odex (I think??), but with trapezoid and midpoint quadrature methods instead of the modified midpoint method.


src/sicmutils/numerical/quadrature/bulirsch_stoer.cljc
```diff
@@ -0,0 +1,230 @@
...
+  ([a b] (h-sequence a b bulirsch-stoer-steps))
+  ([a b n-seq]
+   (map (slice-width a b) n-seq)))
+
+;; ## Bulirsch-Stoer Estimate Sequences
+;;
+;; The next group of functions generates `open-sequence` and `closed-sequence`
+;; methods, analagous to all other quadrature methods in the library.
+
+(defn- fill-defaults
```
littleredcomputer, line 120: It's a nit but I would fold this fn into the point where it is used, since you have used, in extrapolator-fn, the common pattern of having a short and a long fn signature, the latter taking explicit opts

src/sicmutils/numerical/quadrature/bulirsch_stoer.cljc
```diff
@@ -0,0 +1,230 @@
...
+(defn- extrapolator-fn
+  "Allows the user to specify polynomial or rational function extrapolation via
+  the `:bs-extrapolator` option."
+  [opts]
+  (if (= :polynomial (:bs-extrapolator opts))
+    poly/modified-neville
+    rat/modified-bulirsch-stoer))
+
+(defn- bs-sequence-fn
+  "Returns the function body of either `open-sequence` or `closed-sequence` below,
```
littleredcomputer, line 137: It's unclear what "body" means in this context

littleredcomputer: ...ah, I get it, having read ahead now. Still, it's more of an adapter?

littleredcomputer (approved):

PR #127, opened 2020-10-20
"Adaptive" combinator for definite integrators (integrals #8) 

sritchie:
This PR adds an implementation of the "adaptive"-style wrapper that scmutils used in its default integrator. I won't say much here, because the implementation has detailed comments... BUT there are a few key ways that this could be improved to make it really top notch.

One place it fails (where Mathematica does really well) is integration sin(x)/x from 0 => infinity.

Please ask questions for more detail!!


sritchie:
@littleredcomputer , question for you... when we split, should BOTH sides of the adaptive integrator treat the middle point as closed? Or just one of the sides? I copied the approach in scmutils here but maybe there's a bug with that.

src/sicmutils/numerical/quadrature/adaptive.cljc
```diff
@@ -0,0 +1,191 @@
...
+  ## Optional arguments relevant to `adaptive`:
+
+  `:maxterms`: defaults to `*adaptive-maxterms*`. This is passed to the
+  underlying integrators, and determines how long each interval attempts to
+  converge before a subdivision.
+
+  `:adaptive-neighborhood-width`: When dividing an interval, the split point
+  will be within this factor of the midpoint. Set `:adaptive-neighborhood-width`
+  to 0 for deterministic splitting."
+  ([integrator] (adaptive integrator integrator))
```
littleredcomputer, line 132: Have you considered making "Integrator" a kind of object with closed and open entries? Or taking, at the lower level, an enriched interval object with open/closed tagging (so that a primitive integrator takes an arg of the form [:open a :closed b]?

src/sicmutils/numerical/quadrature/adaptive.cljc
```diff
@@ -0,0 +1,191 @@
...
+                          (closed-integrator f l r opts)
+                          (open-integrator f l r opts)))]
+        (loop [stack [[a b (:interval opts)]]
+               sum   (ua/kahan-sum)
+               iteration 0]
+          (if (empty? stack)
+            {:converged? true
+             :iterations iteration
+             :result (first sum)}
+            (let [[l r interval] (peek stack)
```
littleredcomputer, line 149: Can you `let [[[l r interval] & remaining] stack] ...`

littleredcomputer (approved):

src/sicmutils/numerical/quadrature/adaptive.cljc
```diff
@@ -0,0 +1,191 @@
...
+
+;; ## Main Implementation
+;;
+;; The implementation below takes /two/ integrate functions, not the one
+;; described above. This allows us to handle open and closed intervals, instead
+;; of introducing open endpoints at every subdivision. All internal intervals
+;; that don't touch an open endpoint are considered closed.
+
+
+(defn- fill-defaults
```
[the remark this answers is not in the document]
sritchie: Works for me, but let me try and make the case (and please feel free to override):

I pulled this out because it makes it more obvious where to stick new defaults if they pop up; it's not so obvious (imo) when these defaults get merged inside of a function body, and gives us a nice place to make changes if we want to add validations, or specs.

What do you think?

src/sicmutils/numerical/quadrature/adaptive.cljc
sritchie: Yes, I think we should discuss this since I can tell you have some great thoughts on what the interface here should be for validation, and I'm finding myself a little more confused (probably / hopefully this will resolve with some time).

I added some thoughts before in #136 . I was hesitating on pushing this down to the lowest levels because the lowest integrators can't in fact handle open or closed. That always shows up with some level of combination (right?)

src/sicmutils/numerical/quadrature/adaptive.cljc
sritchie: I didn't do this to try to keep a performance-aware template. The destructuring turns the stack into a sequence (it's a vector now); using `peek` and `pop` explicitly, with `conj` to stick the new elements on, gives us that nice vector speed boost.

PR #128, opened 2020-10-20
variable substitution implementations and "improper" combinator (integrals #9)

sritchie:
This PR adds

- a number of variable substitutions that we can use to wrap integrators, and
- `infinite.cljc` with an `improper` combinator.

This ~matches the scmutils implementation of `evaluate-improper-integral`. But I think that function actually has a number of bugs! I didn't do much testing but it looked like integrals from +inf => -inf don't properly negate the result, for example.

We should be good here.


src/sicmutils/numerical/quadrature/substitute.cljc
```diff
@@ -0,0 +1,237 @@
...
+;;
+;; This implementation can only handle one endpoint at a time, and, the way it's
+;; written, both endpoints have to have the same sign. For an easier interface
+;; to this transformation, see `infinite/evaluate-infinite-integral` in
+;; `infinite.cljc`.
+
+(defn infinitize
+  "Performs a variable substitution targeted at turning a single infinite endpoint
+  of an improper integral evaluation an (open) endpoint at 0 by applying the
+  following substitution:
```
eightysteele, line 46: `improper integral evaluation an (open) endpoint at 0` → `improper integral evaluation to an (open) endpoint at 0` (maybe missing a `to` in there).

test/sicmutils/numerical/quadrature/substitute_test.cljc
```diff
@@ -0,0 +1,87 @@
...
+                 :result 0.2193839343960799}
+                ((qs/infinitize qr/open-integral) f 1 ##Inf))
+          "This is a piece of the proper calculation of Euler's constant.")
+
+      (with-comparator (v/within 1e-8)
+        (is (ish? (:result ((qs/infinitize qr/open-integral) f 1 10))
+                  (:result (qr/open-integral f 1 10)))
+            "This is a piece of the proper calculation of Euler's constant.")))))
+
+(deftest power-law-tests
```
littleredcomputer, line 45: How about the volume of [Gabriel's Horn](https://en.wikipedia.org/wiki/Gabriel%27s_Horn) which should be easy

littleredcomputer (approved):

src/sicmutils/numerical/quadrature/substitute.cljc
sritchie: Thanks, fixed!

test/sicmutils/numerical/quadrature/substitute_test.cljc
sritchie:
```
(testing "volume of Gabriel's Horn"
    (let [slice-area (fn [x] (let [y (/ 1 x)]
                              (* Math/PI (* y y))))
          integrate  (qs/infinitize qr/open-integral)]
      (is (ish? {:converged? true
                 :terms-checked 2
                 :result Math/PI}
                (integrate slice-area 1 ##Inf)))))
```

sritchie: (added!)

PR #129, opened 2020-10-20
Final `definite-integral` interface wrapping up all integral work!! (integrals #10)

sritchie:
It's done!! This PR:

- wraps up everything in this series behind a `definite-integral` function and interface
- adds more tests flexing some of the new abilities of the library (most of the abilities are tested by the individual integrators, of course...)

@littleredcomputer, I realize I removed your adaptive simpson's implementation, which is great - wdyt, should we move that over to `simpson.cljc`, and make it its own integrator?

IT'S DONE!


src/sicmutils/numerical/quadrature.cljc
```diff
@@ -0,0 +1,194 @@
...
+;; - An 'integrator' function that matches the interface above for
+;;   `definite-integral` (possibly created with `qc/defintegrator`)
+;;
+;; - a dictionary of extra options. This must contain a `:method` key.
+;;
+;; This latter style is used when the method itself is a specialization of a
+;; more general method.
+
+(def ^:private quadrature-methods
+  {:open                    {:method :adaptive-bulirsch-stoer
```
sritchie, line 66: we've dominated scipy :)

src/sicmutils/numerical/quadrature.cljc
```diff
@@ -0,0 +1,193 @@
...
+                             (qc/infinite? b))
+                       (qi/improper integrate)
+                       integrate)]
+       [integrate (dissoc (merge opts m) :method)]))))
+
+;; ## Final API
+;;
+;; Here we are! The one function you need care about if you're interested in
+;; definite integrals. Learn to use this, and then dig in to the details of
+;; individual methods if you run inton trouble or want to learn more. Enjoy!
```
eightysteele, line 151: `inton` → `into`

sritchie: Thanks!

src/sicmutils/numerical/quadrature.cljc
littleredcomputer: That we have. I really think you have covered some uncharted territory. As an overview, it's clear that GJS wanted strong integrators capable of producing plots of function values with integrals on the RHS, looked at the state of the art, realized that the QUADPACK people had lost the plot, and decided to return to first principles with the Memo. But his approach was depth-first, since numerical analysis is not the focus of his work. This is different, these methods are quite a bit better than just cranking down $h$ and hoping for the best. The empirical measurement of the savings in function evaluations was worth measuring and the Achilles' heel of all the naive techniques...that eventually you will lose significance by subtracting nearly equal things when you lower $h$ too far...is really elegantly captured by all of this work.

test/sicmutils/numerical/quadrature_test.cljc
```diff
@@ -0,0 +1,171 @@
...
+
+      (let [epsilon 1e-9
+            f (partial L epsilon)]
+        (is (near 2.00992 (f 0.15)))
+        (is (near 2.01844 (f 0.30)))
+        (is (near 2.03279 (f 0.45)))
+        (is (near 2.0532 (f 0.60)))
+        (is (near 2.08001 (f 0.75)))
+        (is (near 2.11368 (f 0.9))))))
+
```
littleredcomputer, line 70: Are we getting more significance in the result of this just using straight-up integrate? This one is improper on the left, so by saying that in the options we should get much better results without asking for 1e-9.

littleredcomputer (approved):
Bravo! 🥇

test/sicmutils/numerical/quadrature_test.cljc
sritchie: It's improper on the right, yeah? `0` gives us a denominator of `1 / sqrt(2g (1 - cos(a))`, while `a` gives us a pole.

It looks like we can tighten up the `near` test and remove the options map, and get a passing test:

```clojure
(testing "harder"
    (let [near (v/within 1e-5)
          g 9.8
          integrand (fn [theta0]
                      (fn [theta]
                        (/ (Math/sqrt
                            (* 2 g (- (Math/cos theta)
                                      (Math/cos theta0)))))))
          L (fn [a]
              (let [f (integrand a)]
                (* 4 (q/definite-integral f 0 a {:method :closed-open}))))]
      (is (near 2.00992 (L 0.15)))
      (is (near 2.01844 (L 0.30)))
      (is (near 2.03279 (L 0.45)))
      (is (near 2.0532 (L 0.60)))
      (is (near 2.08001 (L 0.75)))
      (is (near 2.11368 (L 0.9)))))
```

I need to stare at this more, since I've lost touch with the physics and can't remember what system we're describing :)

PR #131, opened 2020-10-27
Bug fix for Brent minimizer, re-enable cljs tests (ODE #1)

sritchie:
This PR:

- fixes a bug in Brent's method... I discovered this when some other examples wouldn't converge, and realized upon closer inspection that this explains one line I had THOUGHT was a bug in the original fortran!
- enables CLJS tests that I'd missed enabling
- disables the 22-minute-long FDG ch7 test in Clojurescript. Once we get the simplifier a little faster we should re-enable; I just wanted some faster turnaround.
- adds an `:all` specifier to the Clojure tests, which are all quite fast now.

This is the first of the final batch of PRs that gets ODEs and function compilation enabled in CLJS!


src/sicmutils/numerical/unimin/brent.cljc
```diff
@@ -119,16 +119,18 @@
...
+    interval `(a, b)`.
 
   - If `delta < tol1`, returns `x + tol1` in the direction of `delta`, to force
     a step of at least `tol1`.
 
   NOTE tol2 == 2*tol1."
   [a x b delta tol1 tol2]
-  (let [near-edge? (or (< (- x a) tol2)
-                       (< (- b x) tol2))]
+  (let [x+delta    (+ x delta)
```
sritchie, line 131: whoops!

src/sicmutils/util/stopwatch.cljc
```diff
@@ -88,6 +88,8 @@
            (Stopwatch. (constantly offset') offset' false))
          this))
 
+     (reset [this] (Stopwatch. nil 0 false))
```
sritchie, line 91: I need to add stopwatch tests. This is a bug for sure.

test/sicmutils/generators.cljc
```diff
@@ -22,9 +22,9 @@
      :clj
      gen/size-bounded-bigint))
 
-#?(:clj
-   (def biginteger
-     (gen/fmap u/biginteger bigint)))
+(def biginteger
```
sritchie, line 25: No real NEED for this but it's nice to have parity in available vars.

test/sicmutils/matrix_test.cljc
```diff
@@ -268,13 +267,13 @@
...
                   (= (g/* (g/* p q) r) (g/* p (g/* q r))))))
 
-;; Both of these are waiting for a fractional implementation, so we don't lose precision.
-#?(:clj
-   (defspec a*ainv=i
-     (gen/let [n (gen/choose 1 5)]
-       (prop/for-all [A (generate-square-matrix n)]
-                     (or (zero? (g/determinant A))
-                         (= (m/I n) (g/* (g/invert A) A) (g/* A (g/invert A))))))))
+(defspec a*ainv=i
```
sritchie, line 270: woohoo!

test/sicmutils/numerical/unimin/brent_test.cljc
```diff
@@ -84,8 +84,8 @@
       (is (ish? {:result (* 3 (/ Math/PI 2))
                  :value -1.0
                  :converged? true
-                 :iterations 30
-                 :fncalls 31}
+                 :iterations 15
```
sritchie, line 87: ANDDD here we go, the effect of the bugfix.

littleredcomputer: Is this still a bug in NR? What was the source of the Fortran?

littleredcomputer (approved):

test/sicmutils/numerical/unimin/brent_test.cljc
sritchie: I studied the code from Brent's book: https://www.amazon.com/Algorithms-Minimization-Without-Derivatives-Mathematics-ebook/dp/B00CRW5ZTK

The Apache Commons and scipy versions are pretty much straight ports. This "bugfix" gets us back to exactly what Brent had in his book, but I missed it because he reassigned to a variable in one conditional branch, and I thought the value was getting ignored. When I stared again I realized that he was in fact using it in this way.

Damnable state!

PR #133, opened 2020-10-29
Added restricted SCI-based compiler + docs, tests (ODE #2)

sritchie:
This PR:

- adds documentation and tests to `sicmutils.numerical.compile`
- changes the default compilation from `eval` to "SCI" https://github.com/borkdude/sci

The huge advantage here is that SCI works with Clojurescript without having to move to self-hosted CLJS. That is apparently a really rare beast, and no one is going to pay that price to use this library.

With SCI support, everything will now work wonderfully from Clojure or CLJS.

If you DO happen to be in a self-hosted CLJS environment, you can use native `eval`. @littleredcomputer , would love your thoughts on whether we should actually keep both or not...

I also did some rewriting of the compilation steps as part of understanding them. Hope this is okay!


src/sicmutils/numerical/compile.cljc
```diff
@@ -18,173 +18,554 @@
...
+(defn- expr-frequencies
+  "Returns a map from distinct subexpressions in the supplied `expr` to the
+  number of times each appears.
+
+  If `skip?` returns true given a subexpression it won't be included as a key in
+  the returned map."
+  [expr skip?]
+  (let [inc* (fnil inc 0)
+        expr->count (atom {})]
+    (w/postwalk (fn [e]
```
sritchie, line 129: I wish I knew how to do this without the atom!

src/sicmutils/numerical/compile.cljc
```diff
@@ -18,173 +18,554 @@
...
+   'acos #(Math/acos %)
+   'asin #(Math/asin %)
+   'atan #(Math/atan %)
+   'expt #(Math/pow %1 %2)
+   'sqrt #(Math/sqrt %)
+   'abs (fn [^double x] (Math/abs x))
+   'log #(Math/log %)
+   'exp #(Math/exp %)})
+
+;; ## Subexpression Elimination
```
sritchie, line 101: hope you like the flow here!

src/sicmutils/numerical/compile.cljc
```diff
@@ -18,173 +18,554 @@
...
+  `body`, given:
+
+  - `params`: a seq of symbols equal in count to the original state function's
+    args
+  - `state-model`: a sequence of variables representing the structure of the
+    nested function returned by the state function
+  - `body`: a function body making use of any symbol in the args above"
+  ([params state-model body]
+   (let [f `(fn ~(state-argv params state-model) ~body)]
+     (sci/eval-form (sci/fork sci-context) f))))
```
sritchie, line 434: how easy is this??

src/sicmutils/numerical/ode.cljc
```diff
@@ -87,7 +87,8 @@
            dimension           (count (flatten initial-state))
            derivative-fn
            (if compile?
-             (c/compile-state-function state-derivative derivative-args initial-state)
+             (let [f' (c/compile-state-fn state-derivative derivative-args initial-state)]
```
sritchie, line 90: there are more changes coming here in #135 . This was the minimal change required to keep it all working for now (and we save a concat!)

Also, in the non-compiled case, we never actually needed to concatenate the derivative-args.

src/sicmutils/numerical/quadrature/infinite.cljc
```diff
@@ -121,7 +121,7 @@
            ab-interval   (qc/interval opts)
            integrate     (partial call integrator)
            inf-integrate (partial call (qs/infinitize integrator))
-           r-break       (Math/abs infinite-breakpoint)
+           r-break       (Math/abs ^double infinite-breakpoint)
```
sritchie, line 124: reflection warning hit me here on compilation.

test/sicmutils/numerical/compile_test.cljc
```diff
@@ -28,51 +29,110 @@
 
 (def ^:private near (v/within 1e-6))
 
-#?(:clj
-   (deftest compile-univariate
-     (let [f (fn [x] (+ 1 (g/square (g/sin x))))
-           cf (c/compile-univariate-function f)]
-       (is (near (f 0.5) (cf 0.5))))))
+(deftest compile-univariate-tests
+  (let [f  (fn [x] (+ 1 (g/square (g/sin x))))
```
sritchie, line 33: didn't add a ton of tests but there are some nice ones below that show multi-level replacements going on.

src/sicmutils/numerical/compile.cljc
sritchie: figured it out :)

src/sicmutils/expression.cljc
```diff
@@ -66,8 +66,9 @@
   "Return the 'variables' (e.g. symbols) found in the expression x,
   which is an unwrapped expression, as a set"
   [x]
-  (if (symbol? x) #{x}
-                  (->> x flatten (filter symbol?) (into #{}))))
+  (if (symbol? x)
+    #{x}
+    (into #{} (filter symbol?) (flatten x))))
```
sritchie, line 71: just a minor update to use the transducer version.

src/sicmutils/numerical/compile.cljc
```diff
@@ -18,173 +18,551 @@
...
+
+(defn- expr-frequencies
+  "Returns a map from distinct subexpressions in the supplied `expr` to the
+  number of times each appears.
+
+  If `skip?` returns true given a subexpression it won't be included as a key in
+  the returned map."
+  [expr skip?]
+  (let [children (partial filter seq?)]
+    (->> (rest
```
sritchie, line 128: I was happy to finally figure out this method. I had extracted this from the code we had before:

```
(defn- expr-frequencies [expr skip?]
  (let [inc* (fnil inc 0)
        expr->count (atom {})]
    (w/postwalk (fn [e]
                  (when (and (seq? e) (not (skip? e)))
                    (swap! expr->count update e inc*))
                  e)
                expr)
    @expr->count))
```

The newer version is just as fast and skips using the atom. Woohoo!

src/sicmutils/mechanics/lagrange.cljc
```diff
@@ -133,7 +133,7 @@
 
 (defn Lagrangian-action
   [L q t1 t2]
-  (q/definite-integral (compose L (Γ q)) t1 t2 {:compile? true}))
+  (q/definite-integral (compose L (Γ q)) t1 t2 {:compile? false}))
```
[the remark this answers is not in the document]
littleredcomputer: It is frustrating. I'd leave the example alone but maybe we should flag this for an updated treatment in the exposition phase of this.

src/sicmutils/numerical/compile.cljc
```diff
@@ -18,173 +18,554 @@
...
+       you wrote it with generic routines)."}
+
+  compiled-fn-whitelist
+  {'up struct/up
+   'down struct/down
+   '+ +
+   '- -
+   '* *
+   '/ /
+   'cos #(Math/cos %)
```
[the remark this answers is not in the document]
littleredcomputer: My plan (which may have been obsoleted by SCI) was to compile functions into RPN, i.e. to a vector of instructions for a JVM-like stack machine where the only opcodes are the supported generic functions. But that rules out things like `if`. I haven't looked at what SCI actually does but I think it's a fine choice for this. I'm on the fence here

src/sicmutils/expression.cljc
littleredcomputer: I had forgotten about them :)

src/sicmutils/numerical/compile.cljc
littleredcomputer: that is cool

littleredcomputer (approved):
this is awesome and now it makes me wonder whether ->JavaScript shouldn't use this technique too, as you hinted.

src/sicmutils/numerical/compile.cljc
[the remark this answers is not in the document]
sritchie: @littleredcomputer yeah, I remember you mentioning this - I THINK right now (please correct me if I'm wrong) we can't have anything but the listed operations because we generate the equation by passing symbols through. So nothing like `if`, a loop etc would survive this process. (right?)

We can of course expand our `generic` vocabulary! JAX https://github.com/google/jax has taken this really far and might be a nice object to study.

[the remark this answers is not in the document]
sritchie: But I selfishly want to study your stack machine implementation so I'm not against it!

PR #134, opened 2020-10-29
Add mutable stopwatch implementation for cljs (ODE #3)

sritchie:
Whoops, looks like immutable isn't the best stopwatch interface :)

I added a mutable wrapper over the original and have made this the default. This all works in Clojure now too! @littleredcomputer , up to you if you want to remove the Guava dependency now that we have this on both sides. What do you think?


src/sicmutils/util/stopwatch.cljc
```diff
@@ -111,13 +56,95 @@
...
+          (+ (elapsed-fn)
+             offset)
+          offset)
+        (from-nanos unit)))
+
+  (repr [this]
+    (let [[x unit] (choose-unit (elapsed this :nanos))]
+      (str x " " (abbreviate unit)))))
+
+(defn- wrapped
```
sritchie, line 122: this was the main change.

littleredcomputer: Sounds good. I "grew up" with Stopwatch, and the thing I like most about it was that it writes µs for microseconds (i.e., whoever wrote it knows what the user wants to see in the output). But yeah, let's kill the Guava dep!

littleredcomputer (approved):

src/sicmutils/util/stopwatch.cljc
sritchie: haha, don't worry, I kept this feature in the Clojure-native version we wrote :)

PR #135, opened 2020-10-29
Odex support in CLJS, enable final CLJS tests + complete conversion! (ODE #4)

sritchie:
This PR:

- Adds a wrapper around Odex
- re-enables all tests that were marked clj-only because of missing features

@littleredcomputer , the interface is not wonderful here, due to the two libraries we're wrapping... but I think it's fine. I expect that we'll massage it as we start using the ODE interface more for animation, now that this is finally done. Woohoo!!


src/sicmutils/examples/central_potential.cljc
```diff
@@ -73,11 +73,11 @@
                           (up xdot_0 ydot_0 0 0))]
     ((e/evolve state-derivative m M)
      initial-state
-     observe
      dt
      t
-     1.0e-6
-     :compile true)))
+     {:compile? true
```
sritchie, line 78: lots of little changes like this. I collected our options into a single map, which imo is a little cleaner (since `observe` feels optional).

src/sicmutils/numerical/ode.cljc
```diff
@@ -120,10 +80,76 @@
...
   ensuring that we also invoke the callback for the final point."
-     [integrator observe step-size initial-state]
-     (.addStepHandler
-      integrator
-      (step-handler observe step-size initial-state))))
+     [^GraggBulirschStoerIntegrator integrator observe step-size initial-state]
+     (let [handler (step-handler observe step-size initial-state)]
+       (.addStepHandler integrator handler))))
+
+(defn integration-opts
```
sritchie, line 87: I am a little conflicted on this namespace @littleredcomputer and would love your thoughts. This function is okay... there is enough in common between what Odex and Apache-Commons need that I'm fine with the multiple internal branches.

What further options should we be allowing in from `opts` to odex?

More questions coming below.

src/sicmutils/numerical/ode.cljc
```diff
@@ -120,10 +80,76 @@
...
+                               (let [d:dt (apply state-derivative derivative-args)
+                                     array->state #(struct/unflatten % initial-state)]
+                                 (comp d:dt array->state))))
+
+        state->array     #?(:clj
+                            (comp double-array flatten)
+
+                            :cljs
+                            (fn [state]
+                              (->> (flatten state)
```
sritchie, line 115: We may be able to save the `map u/double` in the compiled case, but not if we don't compile, since ratios, js/BigInt etc can bubble up, and these break Odex when they get in.

src/sicmutils/numerical/ode.cljc
```diff
@@ -120,10 +80,76 @@
...
+                                (let [y' (-> (derivative-fn y)
+                                             (state->array)
+                                             (doubles))]
+                                  (System/arraycopy y' 0 out 0 (alength y')))
+                                (us/stop evaluation-time))
+                              (getDimension [_] dimension))
+
+                            :cljs
+                            (fn [_ y]
+                              (us/start evaluation-time)
```
sritchie, line 133: I wonder if these timers and counters are something the user should add themselves if they want them. We have a `u/counted` function wrapper. `u/timed` would make great sense too. Then we could strip them out if we like... what do you think?

src/sicmutils/numerical/ode.cljc
```diff
@@ -120,10 +80,76 @@
...
+                      (GraggBulirschStoerIntegrator. 0. 1.
+                                                     (double epsilon)
+                                                     (double epsilon))
+
+                      :cljs
+                      (let [solver (o/Solver. dimension)]
+                        (set! (.-absoluteTolerance solver) epsilon)
+                        (set! (.-relativeTolerance solver) epsilon)
+                        solver))]
+    {:integrator integrator
```
sritchie, line 148: these aren't quite isolated enough since you still have to know what type you're dealing with to do anything with the return values. I didn't want to make a protocol yet, since maybe we'll have a Clojure implementation that's fast and can use the same code for both branches.

src/sicmutils/numerical/ode.cljc
```diff
@@ -137,46 +163,85 @@
...
+               (integration-opts state-derivative derivative-args initial-state opts)
+               initial-state-array (into-array
+                                    (flatten initial-state))
+               array->state #(struct/unflatten % initial-state)
+               output-buffer (double-array dimension)
+               observe-fn    (when observe
+                               (set! (.-denseOutput integrator) true)
+                               (.grid integrator step-size
+                                      (fn [t y]
+                                        (reset! latest t)
```
sritchie, line 180: @littleredcomputer , note for you. This MIGHT be an odex bug.

I found that with `denseOutput`, if I had a grid like `0 -> 1`, measured every `0.1`, the imprecise double arithmetic meant that my final point was around `0.99999999999` or something. Since the endpoint was basically included, tacking it on at the end gave me one more point with Odex than with the Java implementation.

For a `dt` that doesn't evenly divide into the range I always matched Java when I manually observed the final value, as I do here.

My compromise for now is to track the latest timestamp and use `near?`, as you see here, to avoid observing the final value if it's very near the last value logged by Odex.

Does that make sense? Any thoughts on how to avoid this?

src/sicmutils/numerical/ode.cljc
```diff
@@ -137,46 +163,85 @@
...
                array->state #(struct/unflatten % initial-state)
                output-buffer (double-array dimension)]
            (when observe
              (attach-handler integrator observe step-size initial-state))
-           (.integrate integrator equations 0
+           (.integrate ^GraggBulirschStoerIntegrator
+                       integrator equations 0
                        initial-state-array t output-buffer)
-           (doto total-time (us/stop) (us/reset))
+           (us/stop total-time)
```
sritchie, line 207: whoops, I was resetting before logging!

test/sicmutils/examples/central_potential_test.cljc
```diff
@@ -177,10 +177,7 @@
                             (* (expt y 2) (sqrt (+ (expt X 2) (* -2 X x) (expt Y 2) (* -2 Y y) (expt x 2) (expt y 2))))))))
              (e/simplify ((central/state-derivative 'm 'M) state))))
 
-      #?(:clj
-         ;; Integrator isn't yet implemented in cljs.
```
sritchie, line 181: it all works now in cljs!

test/sicmutils/sicm/ch1_test.cljc
```diff
@@ -91,42 +91,34 @@
...
-         (let [m (e/minimize
-                  (varied-free-particle-action 3.0 test-path (up sin cos square) 0.0 10.0)
-                  -2.0 1.0)]
-           (is (near 0.0 (:result m)))
-           (is (near 435 (:value m)))))
-
-      #?(:clj
-         ;; TODO activate when we get integrator support in cljs.
-         (is (near 436.2912143 ((varied-free-particle-action 3.0 test-path (up sin cos square) 0.0 10.0) 0.001))))
+      (is (= 435.0 (e/Lagrangian-action (L/L-free-particle 3.0) test-path 0.0 10.0)))
```
sritchie, line 94: YES, finally!! Feels great to activate these.

src/sicmutils/numerical/ode.cljc
littleredcomputer: count on it

src/sicmutils/numerical/ode.cljc
littleredcomputer: I agree, it's the user's choice. My feeling earlier was that somebody who was willing to rev up this much symbolic and numerical computational machinery could tolerate some logging. From where I come from, logging is essential. In the math world, it's uncommon, and you can't make everybody happy. For the present I would leave it in there.

src/sicmutils/numerical/ode.cljc
littleredcomputer: I'd have to think about that but the main thing, I think, is that this should all become pure Clojure, which, thanks to your efforts, is now on the horizon.

src/sicmutils/numerical/ode.cljc
littleredcomputer: I would say let the 0.999999 through...on the principle that, that is in fact what ODEX did for the integration. Almost any practical application would be unaffected by our not returning the exact double 1.0 here, so I'd say, if you can't tolerate this, find something better than ODEX for the integration :)

src/sicmutils/numerical/ode.cljc
littleredcomputer: Are you proposing that the integrators could work entirely in (say) rational arithmetic?

littleredcomputer (approved):
Yup, this does feel good.

src/sicmutils/numerical/ode.cljc
sritchie: I agree too that it's essential here, and extremely helpful to see the drop in times, etc when you turn compilation on. I don't want to go down the road of stripping all logging out and forcing everyone to wire in their own hooks everywhere.

I found out recently that Clojurescript supports Google Closure's ability to set compile time variables, so if someone really cared we could wire up the logging calls to get stripped out with a particular flag for someone's "production build".

I think your suggestion is the right one!

src/sicmutils/numerical/ode.cljc
sritchie: Ah, sorry, that of course is interesting BUT what I meant here was that if you want to build on top of this function, as I did long ago playing with the first animations:

https://github.com/sritchie/maturin/blob/master/src/maturin/better.clj#L43

you still have to know that on the JVM you call

```clojure
 (.integrate integrator equations time s t2 buffer)
```

and in JS you call

```clojure
(.solve integrator equations 0 initial-state-array t observe-fn)
```

No big deal, I was just noting that if we had a `-solve` protocol method we could hide the details.

src/sicmutils/numerical/ode.cljc
sritchie: Ah, yes, I absolutely agree we need to return it; I should have been more clear with what I was guarding against.

In our `observe` treatment, on the JVM, we use the interpolator stuff that the Commons implementation returns to check our intermediate values, and then explicitly pass the final value in to `observe`.

The guard I added here was to make sure that don't do this explicit final observation ourselves of `1.0`, say, when Odex has already observed `0.9999999`.

Without the guard, `observe` would see values for `0.9999999` AND `1.0`. (I'll say it another way since this is reading as quite confusing:

- when `dt` does NOT evenly divide, we have to manually observe the final value, and
- when `dt` DOES evenly divide, we want to skip doing this.

Feels a little wonky to do it with the atom, but maybe it's fine...

PR #141, opened 2020-10-30
Remove series/take (clojure's take works great!)

sritchie:
Since a series is a proper sequence, `take` by itself does a great job here. We can make examples simpler, and get the ability to use the transducer form of `take` with series:

```clojure
(let [render (comp ->infix simplify)
      series (((exp (* 'h D)) tan) 0)]
  (render
   (transduce (take 10) + series)))
;; "62/2835 h⁹ + 17/315 h⁷ + 2/15 h⁵ + 1/3 h³ + h"
```


littleredcomputer (approved):
how did I miss this

sritchie:
@littleredcomputer I can't remember if we had all the seq methods implemented before... but I was able to slim it down even a little more. Thanks!

PR #142, opened 2020-11-01
Power series overhaul (series #2)

sritchie:
This PR upgrades `series.cljc` based on the new Power Series implementation introduced in #152.

Some highlights:

- The matrix data type gains `g/exp` and all of the trig operators (they're implemented as taylor series expansions of those fns)
- operator gains all of the trig operators, and a tidier implementation of `g/exp`
- the calculus `taylor-series` function shrinks down to, basically, `(g/exp D)`
- series has a bunch of new generic methods, all of the trig methods, g/exp, and the ability to functional composition, reversion, all sorts of awesome stuff!

Here's what I added to the CHANGELOG:

### Power Serious, Power Serious

The Power Series implementation in `series.cljc` received an overhaul. The
implementation now follows Doug McIlroy's beautiful paper, ["Power Series, Power
Serious"](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.333.3156&rep=rep1&type=pdf).
Doug also has a 10-line version in Haskell on [his
website](https://www.cs.dartmouth.edu/~doug/powser.html).

The API now offers two types:

 - `Series`, which represents a generic infinite series of arbitrary values, and
 - `PowerSeries`, a series that represents a power series in a single
   variable; in other words, a series where the nth entry is interpreted as
   the coefficient of $x^n$:

    $$[a b c d ...] == $a + bx + cx^2 + dx^3 + ...$$

`series/series?` responds true to both. `series/power-series?` only responds
true to a `PowerSeries`.

To turn a `PowerSeries` into a `Series`, call it as a function with a single
argument, or pass the series and one argument to `series/value` to evaluate the
series using the above equation.

To turn a `Series` into a `PowerSeries`, call `series/->function`. None of the
functions discussed below can mix series types; you'll have to do the conversion
explicitly.

Each type supports the following generic operations:

- `*`, `+`, `-`, `/` between series and non-series
- `g/negate`, `g/invert`, `g/sqrt`, `g/expt` work as expected.
- `g/add` between series and non-series

`PowerSeries` additionally supports:

- `g/exp`, `g/cos`, `g/sin`, `g/asin`, `g/tan`
- `g/partial-derivative`, so `PowerSeries` works well with `D`

Each of these acts as function composition for the single variable function that
the `PowerSeries` represents. If `s` is a `PowerSeries` that applies as `(s x)`,
`(g/exp s)` returns a series that represents `(g/exp (s x))`.

There are many more new methods (see the namespace for full documentation):

- `starting-with` renamed to `series`
- `power-series`, analogous `series` but generates a `PowerSeries`
- `series*` and `power-series*` let you pass an explicit sequence
- `series/take` removed in favor of `clojure.core/take`, since both series
  objects act as sequences
- `generate` takes an additional optional argument to distinguish between series
  and power series
- `Series` now implements more of `v/Value`
- new `zero`, `one`, `identity` constants
- `constant` returns a constant power series
- `xpow` generates a series representing a bare power of `x`
- `->function` turns a `Series` into a `PowerSeries`
- `value`, `fmap` now handles both `Series` and `PowerSeries`
- `(inflate s n)` expands each term $x^i$ of `s` to $x^{in}$
- `compose` returns the functional composition of two `PowerSeries`
- `revert` returns the functional inverse of two `PowerSeries`
- `integral` returns a series representing the definite integral of the supplied
  `PowerSeries`, 0 => infinity (optionally takes an integration constant)

The namespace also provides many built-in `PowerSeries` instances:

- `exp-series`
- `sin-series`
- `cos-series`
- `tan-series`
- `sec-series`
- `asin-series`
- `acos-series`
- `atan-series`
- `acot-series`
- `sinh-series`
- `cosh-series`
- `tanh-series`
- `asinh-series`
- `atanh-series`
- `log1+x-series`
- `log1-x-series`
- `binomial-series`

And two `Series` (non-power):

- `fib-series`, the fibonacci sequence
- `catalan-series`, the [Catalan
  numbers](https://en.wikipedia.org/wiki/Catalan_number)

### Matrix Generic Operations

`::matrix` gained implementations for `exp`, `cos`, `sin`, `asin`, `tan`,
`acos`, `asin`; these now return taylor series expansions of the operator, where
multiplication is composition as before.

### Operator Generics

`Operator` gained implementations for `cos`, `sin`, `asin`, `tan`, `acos`,
`asin`; these now return taylor series expansions of the operator, where
multiplication is composition as before.


src/sicmutils/series.cljc
```diff
@@ -18,24 +18,40 @@
...
-     (:import (clojure.lang AFn IFn Sequential Seqable))))
+     (:import (clojure.lang AFn IFn
+                            Seqable ISeq))))
 
-;; We would prefer to just use native Clojure lazy sequences to represent
-;; series objects. But, they must be invokable as functions, so we must
-;; wrap them in a deftype.
+;; # Power Series
+;;
+;; Following Power Series here!
```
littleredcomputer, line 32: Did you want to say "Serious" here

src/sicmutils/series.cljc
```diff
@@ -89,10 +106,17 @@
     (applyTo [s xs] (AFn/applyToHelper s xs))
 
     Seqable
-    (seq [_] s)]
+    (seq [_] s)
```
littleredcomputer, line 109: Do we need them both? Memory fades. Was it implementing seqable instead of iseq that made the now-retired `series/take` necessary?

src/sicmutils/series.cljc
```diff
@@ -143,108 +167,293 @@
...
 
+(defn- starting-with*
+  "Version that lets us specify arities"
+  ([prefix]
+   (starting-with* prefix v/arity:exactly-0))
+  ([[x :as prefix] arity]
+   (let [s (lazy-cat prefix (repeat (v/zero-like x)))]
+     (->Series arity s))))
+
 (defn starting-with
```
littleredcomputer, line 194: I wonder why I didn't just "copy" the arity of the supplied element

littleredcomputer (approved):
I'm all for getting rid of `::coseries`. I recall getting to understand `cofunction`: that was a signal for the pointwise approach. Maybe we don't need it here. (I.e., adding a function and a function is not the same as adding a function and a cofunction, but it could be that the stronger multidispatch type system we have here makes this easier to do

src/sicmutils/series.cljc
sritchie: :) yup

src/sicmutils/series.cljc
sritchie: `Seqable` actually did this! I added the `ISeq` so I could implement `cons`, so cons-ing onto a series keeps it a series. You're probably right that this could go away now. I'll poke around in the guts.

src/sicmutils/series.cljc
sritchie: I've been studying this, and it looks like they're really using this to mean "series" vs "power series". I was thinking of just making two separate types.

Right now, we only have the "series" implemented; the "power series" version can take a single variable (they don't have multivariable power series) and evaluate it at some indeterminate.

sritchie:
@littleredcomputer , this isn't done yet, but I implemented everything for sequences and I think you'll really like the result!

src/sicmutils/series.cljc
sritchie: Okay, `ISequential` mattered to get `nth` working, otherwise `Seqable` was enough.

src/sicmutils/series.cljc
sritchie: I implemented this and I'm convinced that GJS was using arity as a hack here, and only these two values are allowed. I bet he was trying to open the door for multivariate power series down the road, but I think not using `arity 0` for the applied-series case is clearer.

sritchie:
@littleredcomputer , I know you approved this, but you may want to peek again - I'll merge, but I added lots of tests and docs, and summarized everything in the changelog description. Let me know what you think!

PR #150, opened 2020-11-06
Small fixes, default expt, sub, div implementations

sritchie:
This PR:

- adds a default implementation of `expt` based on a valid `mul`, `kind` and `one-like` implementation
- adds a default `sub`, based on `add` and `negate`
- default `div` if `mul` and `invert` are present

I was inspired by Doug McIlroy's gentle boast that these come for free in Haskell, and realized that we should make it easier for folks to implement these methods too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
-    "scripts": {
-        "test": "shadow-cljs compile test && node target/main/node-tests.js;",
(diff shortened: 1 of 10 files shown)


sritchie:
@littleredcomputer , I’ll merge this since this was part of the #142 approved PR. Next comes that code in a larger block.

PR #155, opened 2020-11-13
Upgrade pattern matcher, simplifier [in progress]

sritchie:
I'm taking on an essay-ification of the rules engine and term-rewriting system! The goal is to get `new-simplifier` implemented.

GJS has taken down the notes that the matcher code was based on... but it turns out that Alexey published this great MIT scheme library, which is a nicely documented, more-fully-fleshed-out version of the code that scmutils integrated: https://github.com/axch/rules

There is a unifier-rule-simplifier in scmutils... but that is not what's active now. use `rule-simplifier.scm`.

Some notes I'd kept on prior art:
- https://github.com/kovasb/combinator has some smoking fast term rewriting in Clojure, 2x faster than Mathematica in Clojurescript. This will be a huge boost. This does NOT say anything about the actual rules, just the process of applying them to the expression tree.
- https://github.com/clojure-numerics/expresso references `combinator`, and is also a nice symbolic arithmetic system.
- https://github.com/mschuene/stratege notes that Expresso and friends are written on top of core.logic and can overflow. We don't do this, so probably doesn't apply. But worth reading another take.
- nice timing, a term rewriting paper from Stephen Diehl: https://www.stephendiehl.com/posts/exotic02.html

- [ ] in rules-syntax in scmutils, looks like he supports a 2-elem and 3-elem rule. What's the diff? There are examples of the 2 element rules in the big one I need, `sqrt-expand` and `sqrt-contract`.


littleredcomputer (approved):

sritchie:
Closing in favor fo #164

PR #159, opened 2020-11-15
Revert "Add more square-root rules to simplifier, bugfix"

sritchie:
Reverts littleredcomputer/sicmutils#158 - @littleredcomputer , I realized I had misread, I think that this is not a correct property for combing radicals.

```
(sqrt(5/2) + sqrt(7/2)) = - sqrt((5 + 7) / 2) == 3.45197
```

```
 sqrt((5 + 7) / 2) == 2.44949
```


littleredcomputer:
Yikes! I had this vague sense that I had overlooked something while I slept. Let me try again.

PR #160, opened 2020-11-15
more sqrt simplifications

sritchie:
This PR adds more `sqrt` simplification rules to allow square roots to cancel out across a division boundary, with or without products in the numerator and denominator.

### DIFF
--- a/src/sicmutils/rules.cljc
+++ b/src/sicmutils/rules.cljc
@@ -80,11 +80,44 @@
     (sqrt (expt :x (:? n even-integer?)))
     => (expt :x (:? #(/ (% 'n) 2)))
 
+    (expt (sqrt :x) (:? n odd-integer?))
+    => (* (sqrt :x) (expt :x (:? #(/ (- (% 'n) 1) 2))))
+
+    (/ :x (sqrt :x)) => (sqrt :x)
+
+    (/ (sqrt :x) :x) => (/ 1 (sqrt :x))
+
+    (/ (* :u* :x :v*) (sqrt :x))
+    =>
+    (* :u* (sqrt :x) :v*)
+
+    (/ (* :u* (sqrt :x) :v*) :x)
+    =>
+    (/ (* :u* :v*) (sqrt :x))
+
+    (/ :x (* :u* (sqrt :x) :v*))
+    =>
+    (/ (sqrt :x) (* :u* :v*))
+
+    (/ (sqrt :x) (* :u* :x :v*))
+    =>
+    (/ 1 (* :u* (sqrt :x) :v*))
+
+    (/ (* :p* :x :q*)
+       (* :u* (sqrt :x) :v*))
+    =>
+    (/ (* :p* (sqrt :x) :q*)
+       (* :u* :v*))
+
+    (/ (* :p* (sqrt :x) :q*)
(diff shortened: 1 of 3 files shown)


src/sicmutils/rules.cljc
```diff
@@ -80,11 +80,44 @@
...
+
+    (/ :x (sqrt :x)) => (sqrt :x)
+
+    (/ (sqrt :x) :x) => (/ 1 (sqrt :x))
+
+    (/ (* :u* :x :v*) (sqrt :x))
+    =>
+    (* :u* (sqrt :x) :v*)
+
+    (/ (* :u* (sqrt :x) :v*) :x)
```
littleredcomputer, line 94: Does this one make its companion above redundant too?

src/sicmutils/rules.cljc
```diff
@@ -80,11 +80,44 @@
...
+
+    (/ :x (* :u* (sqrt :x) :v*))
+    =>
+    (/ (sqrt :x) (* :u* :v*))
+
+    (/ (sqrt :x) (* :u* :x :v*))
+    =>
+    (/ 1 (* :u* (sqrt :x) :v*))
+
+    (/ (* :p* :x :q*)
```
littleredcomputer, line 106: Isn't there a rule above that accomplishes this in a simpler case (so that this case alone would be effective)?
I have been in lazy-evaluation mode when it comes to importing scmutils simplification rules since it takes a patient eye to see what they are doing. Now, if we had evidence that the simpler rules acted as a short circuit that would speed things up, that would be one thing. Easy enough to test to see if fewer rules would do the job?

src/sicmutils/rules.cljc
```diff
@@ -80,11 +80,44 @@
     (sqrt (expt :x (:? n even-integer?)))
     => (expt :x (:? #(/ (% 'n) 2)))
 
+    (expt (sqrt :x) (:? n odd-integer?))
+    => (* (sqrt :x) (expt :x (:? #(/ (- (% 'n) 1) 2))))
+
+    (/ :x (sqrt :x)) => (sqrt :x)
```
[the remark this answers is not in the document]
sritchie: It might be (you know more here!) if there were some mechanism that knew that `(expt :x 1)` was equivalent... but the tests fail to transform that case without this explicit rule. wdyt?

src/sicmutils/rules.cljc
sritchie: Definitely help me through this if I'm missing something... these FEEL quite redundant, but I think need to be here to cover the cases of:

- square root on top or bottom

crossed with

- product in numerator
- product in denominator
- product in both

right?

src/sicmutils/rules.cljc
sritchie: I think with this rule alone (and its companion), everything would totally work if the system was smart enough to wrap `:x` into `(* x)`... but without that we need the other cases, to distinguish the presence of the product. Is that right? Just learning here myself, first dig into this stuff!

littleredcomputer: On second look I think you're right. It has been too long since I worked with these.

littleredcomputer (approved):

PR #164, opened 2020-11-17
Upgrade pattern matcher, simplifier, add rules [in progress]

sritchie:


littleredcomputer:
What is the rationale for implementing new-simplifier? Do you want to be able to import the simplification sequence as it stands in sum?

src/sicmutils/rules.cljc
```diff
@@ -189,70 +235,228 @@
...
+   ))
+
+(def log-extra
+  (ruleset
+   (* (:? n v/integral?) :f1* (log :x) :f2*)
+   =>
+   (* :f1* (log (expt :x :n)) :f2*)
+   ))
+
+(def canonicalize-partials
```
sritchie, line 332: I added this in #286 , remove it here.

src/sicmutils/rules.cljc
```diff
@@ -189,70 +235,228 @@
...
+                                 (eq? (car thing) 'down))))
+                    :else false))])
+    (if (= (length args) 1)
+      (sea access-chain (first args))
+      (sea (rest access-chain)
+           (nth args (first access-chain))))))
 
-(def ^:private flush-obvious-ones
+;; TODO taken from scheme.
+(def canonicalize-partials*
```
sritchie, line 371: whoops, this is what I'd added!

breandan:
It might also be interesting to compare with Tate et al.'s work on [equality saturation](https://www.cs.cornell.edu/~ross/publications/eqsat/). Not sure if there is a JVM implementation of [egg](https://egraphs-good.github.io/), but I saw @erdos has some [experiments in Clojure](https://github.com/erdos/cogent). cc: @mwillsey

sritchie:
Closing in favor of #353 .

PR #165, opened 2020-11-20
Fix broken link in README

zane:
The "apply hooks" link in the "Background" section of the `README` triggers a `404`. It appears that the new location for that content is [here](https://www.gnu.org/software/mit-scheme/documentation/stable/mit-scheme-ref/Application-Hooks.html). This pull request updates the link accordingly.


sritchie:
Thank you @zane !

PR #166, opened 2020-11-24
Speed up power series multiplication

sritchie:
I was playing around to see if I could speed this up by generating the binomial series directly in the exponent code... and it looks like no, it's plenty fast now with a `zero?` check added in the mult method.


littleredcomputer (commented):
LGTM

PR #168, opened 2020-11-29
[monster pr] split apart core numeric tower and symbolic expression path

sritchie:
## Numerics

The MAJOR thing this PR accomplishes is that numbers no longer inherit from `::x/numerical-expression`. I traced my way through the Scheme, and I think the way they build their stack is nice enough that we should use it here.

Numbers (in scmutils, and now here) implement all generics directly, so calling something like `(cos 12)` will always hit `(Math/cos 12)`. Before this change, that call would run through the implementations in `numsymb.cljc`. This was a problem with default implementations of `sec`, say, where calling `(sec 12)` would return a symbolic `(/ 1 (cos 12))`.

So now:

- numbers are their own thing, inheriting from `::v/number`.
- `Complex` now extends `::v/number`, so complex numbers, just lilke in Scheme return `true` for `number?` checks
- we have a new `::v/real` type that rationals, floating pt and all the rest extend from.
- `v/numerical?` is the ACTUAL test we want to do to see if something is either an expression representing a number or an actual number.

All of the code in `numsymb.cljc` is "number/symbol` because if you CHOOSE, you can wrap an actual number in `(literal-number 12)`, say, and it will hit all of the good stuff in `numsymb.cljc`. That's why these methods handle both.

## Complex Generics

I've moved:

- `real-part`
- `imag-part`
- `angle`
- `conjugate`

from `complex.cljc` to `generic.cljc` and made these proper extensible generics. They work complex, real numbers and symbolic expressions.

## Expressions

- changed `Expression` to `Literal`
- renamed `:sicmutils.expression/numerical-expression` to `::x/numeric`
- `x/walk-expression` renamed to `x/evaluate`
- moved the code that installs `::x/numeric` into the generic system into `sicmutils.abstract.number`. My thought here is that this will be a sane place for us to add abstract matrices and structures, now that it's clearer for me how to do that.

I also added some slight upgrades to our expression walking code:

We have a `substitute` method now, as well as a `compare` function that returns a comparator for expressions that works the way the `scmutils` one did. (Actually better... GJS implements `<`, but it was a little wonky. For example, `(< complex-number expr)` and `(< expr complex-number)` are both true, but they're of course not equal. That's fixed here.

### TODO

- [x] test that complex numbers return true to `number?`
- [x] make `variables-in` return an empty set if you pass it a number
- [x] many more tests...


src/sicmutils/abstract/number.cljc
```diff
@@ -0,0 +1,118 @@
...
+            [sicmutils.expression :as x]
+            [sicmutils.generic :as g]
+            [sicmutils.numsymb :as ns]
+            [sicmutils.simplify :as s]
+            [sicmutils.util :as u]
+            [sicmutils.value :as v])
+  #?(:clj
+     (:import (clojure.lang Symbol))))
+
+(defn literal-number [x]
```
sritchie, line 31: @littleredcomputer , I'll add documentation and tests for everything here, but me explain a bit now. This is now the blessed way to create a "literal" number; these have the property that when you do binary operations between these and a normal number, or anything else that they're compatible with, the other thing gets absorbed into the "expression".

So if you want to keep a tree of exact arithmetic, you can do `(literal-number 12)`, and operations will keep the result exact:

```clojure
sicmutils.abstract.number> (g/+ 1 (g/cos (literal-number 12)))
(+ 1 (cos 12))
```

they won't try to do this if the number is NOT exact:

```clojure
(g/+ 1 (g/cos (literal-number 12.2)))
1.9336336440746373
```

src/sicmutils/calculus/derivative.cljc
```diff
@@ -47,7 +47,7 @@
   (one-like [_] 1)
   (freeze [_] `[~'Differential ~@terms])
   (exact? [_] false)
-  (numerical? [d] (g/numerical-quantity? (differential-of d)))
```
sritchie, line 50: okay, this all makes way more sense to me now! We can use `v/numerical?` everywhere and don't need this extra wrapper.

src/sicmutils/complex.cljc
```diff
@@ -43,19 +43,20 @@
   [a]
   (instance? Complex a))
 
-(defn conjugate [^Complex a] (.conjugate a))
-(defn real-part [^Complex a] (#?(:clj .getReal :cljs .-re) a))
-(defn imag-part [^Complex a] (#?(:clj .getImaginary :cljs .-im) a))
-(defn angle [^Complex a] (#?(:clj .getArgument :cljs .arg) a))
+(defmethod g/real-part [::complex] [^Complex a] (#?(:clj .getReal :cljs .-re) a))
```
sritchie, line 46: these are now exposed as extensible generics.

src/sicmutils/complex.cljc
```diff
@@ -84,19 +85,23 @@
...
+                    im (g/imag-part c)]
+                (if (v/nullity? im)
+                  re
+                  (list 'complex re im))))
+  (exact? [c] (and (v/exact? (g/real-part c))
+                   (v/exact? (g/imag-part c))))
   (numerical? [_] true)
   (kind [_] ::complex))
 
 (defmethod g/add [::complex ::complex] [^Complex a ^Complex b] (.add a b))
```
sritchie, line 98: you'll see as you go through that `v/numtype == ::v/number` is now `::v/real`. `::v/number` still exists, but now complex numbers inherit from it. `::v/real` is what the old thing meant.

src/sicmutils/expression.cljc
```diff
@@ -18,70 +18,204 @@
...
+    ::vector
+    ::abstract-down
+    ::abstract-matrix})
 
+(deftype Literal [type expression meta]
   v/Value
-  (nullity? [_] false)  ;; XXX what if it's a wrapped zero? one?
-  (unity? [_] false)
+  (nullity? [_]
+    (and (v/number? expression)
```
sritchie, line 38: woohoo, todo covered!

src/sicmutils/expression.cljc
```diff
@@ -18,70 +18,204 @@
...
+  (unity? [_]
+    (and (v/number? expression)
+         (v/unity? expression)))
+
   (zero-like [_] 0)
   (one-like [_] 1)
-  (numerical? [_] (= type ::numerical-expression))
-  (exact? [_] false)
+  (numerical? [_] (= type ::numeric))
+  (exact? [_] (and (v/number? expression)
```
sritchie, line 48: same here, we stay exact if we wrap an exact number.

src/sicmutils/expression.cljc
```diff
@@ -18,70 +18,204 @@
...
+                   (v/exact? expression)))
   (freeze [_] (v/freeze expression))
-  (kind [_] type))
+  (kind [_] type)
+
+  Object
+  (toString [_] (str expression))
+  #?(:clj
+     (equals [a b]
```
sritchie, line 56: This is what forced the `defrecord` => `deftype` move - we want to override equality to use `v/eq` so we can compare to numbers etc.

src/sicmutils/expression.cljc
```diff
@@ -18,70 +18,204 @@
...
 
 #?(:clj
-   (defmethod print-method Expression [^Expression s ^java.io.Writer w]
+   (defmethod print-method Literal [^Literal s ^java.io.Writer w]
      (.write w (.toString s))))
 
-(defn literal-number
-  [expression]
-  (if (number? expression)
-    expression
```
sritchie, line 45: it used to be that `literal-number` would act as identity for numbers. Not anymore! Also not in `scmutils`.

src/sicmutils/expression.cljc
```diff
@@ -18,70 +18,204 @@
...
+  "Applies f to the expression part of e and creates from that a Literal
+  otherwise like e."
+  [f ^Literal e]
+  (->Literal (.-type e)
+             (f (.-expression e))
+             (.-meta e)))
+
+;; ## Metadata
+
+(defn properties [x]
```
sritchie, line 113: we don't use this yet, but I think we should keep it on now (tidied and tested of course). GJS uses this to add properties like `:real`, `:one`, `:zero` etc to abstract matrices.

What do you think, @littleredcomputer ?

src/sicmutils/expression.cljc
```diff
@@ -18,70 +18,204 @@
...
+;; ## Expression Walking
 
 (defn variables-in
   "Return the 'variables' (e.g. symbols) found in the expression x,
-  which is an unwrapped expression, as a set"
-  [x]
-  (if (symbol? x)
-    #{x}
-    (into #{} (filter symbol?) (flatten x))))
+  which is a wrapped or unwrapped expression, as a set"
```
sritchie, line 144: modified this so it will do the right thing if you give it a wrapped expression too.

src/sicmutils/expression.cljc
```diff
@@ -18,70 +18,204 @@
...
+  association is used for elements in function application position (first of a
+  sequence)."
+  [expr sym->var sym->f]
+  (letfn [(walk [node]
+            (cond (symbol? node) (sym->var node node)
+                  (sequential? node)
+                  (let [[f-sym & args] node]
+                    (if-let [f (sym->f f-sym)]
+                      (apply f (map walk args))
+                      (u/illegal (str "Missing fn for symbol - " f-sym))))
```
sritchie, line 162: added a better error message here, explaining what the problem is

src/sicmutils/expression/analyze.cljc
```diff
@@ -17,11 +17,11 @@
 ; along with thixs code; if not, see <http://www.gnu.org/licenses/>.
 ;
 
-(ns sicmutils.analyze
```
sritchie, line 20: I've moved this file to `sicmutils.expression.analyze`.

src/sicmutils/function.cljc
```diff
@@ -233,7 +236,7 @@
 (make-binary-operation g/sub g/-)
 (make-binary-operation g/mul g/*)
 (make-binary-operation g/div g/divide)
-(make-binary-operation g/expt g/expt)
```
sritchie, line 236: added an arity where you only have to supply one.

src/sicmutils/function.cljc
```diff
@@ -243,20 +246,36 @@
...
+
 (make-unary-operation g/sinh)
 (make-unary-operation g/cosh)
 (make-unary-operation g/tanh)
 (make-unary-operation g/square)
 (make-unary-operation g/cube)
 (make-unary-operation g/exp)
 (make-unary-operation g/log)
-(make-unary-operation g/transpose)
+(defmethod g/transpose [::function] [f]
```
sritchie, line 260: @littleredcomputer , this comes from scmutils:

```scheme
;;; This only makes sense for linear functions...
(define (((f:transpose f) g) a)
  (g (f a)))
```

I was going to ask if you know what the deal is here?

src/sicmutils/function.cljc
```diff
@@ -243,20 +246,36 @@
...
+  (fn [g]
+    (fn [a]
+      (g (f a)))))
+
+(make-binary-operation g/cross-product)
+(make-binary-operation g/gcd)
+
+;; Complex Operations
 
-;; TODO sinh cosh ...
```
sritchie, line 256: whoops, dangling TODO!

src/sicmutils/function.cljc
```diff
@@ -243,20 +246,36 @@
...
+(make-unary-operation g/real-part)
```
sritchie, line 270: these are extended here just like in `scmutils`.

src/sicmutils/function.cljc
```diff
@@ -243,20 +246,36 @@
...
+(make-unary-operation g/imag-part)
+(make-unary-operation g/magnitude)
+(make-unary-operation g/angle)
+(make-unary-operation g/conjugate)
 
 (defmethod g/simplify [Function] [a] (g/simplify (:name a)))
-(derive ::x/numerical-expression ::cofunction)
+(derive ::x/numeric ::cofunction)
+(derive ::v/number ::cofunction)
```
sritchie, line 278: for now, before I properly think about what `::cofunction` means etc, I've just added `derive ::v/number` everywhere I had `::x/numeric` alone before.

src/sicmutils/generic.cljc
```diff
@@ -21,31 +21,10 @@
...
   (:require [sicmutils.value :as v]
-            [sicmutils.expression :as x]
             [sicmutils.util :as u])
   #?(:cljs (:require-macros [sicmutils.generic :refer [def-generic-function]]))
   #?(:clj
-     (:import [clojure.lang LazySeq PersistentVector Symbol Seqable Var])))
-
-;;; classifiers
-
-(defn literal-number?
```
sritchie, line 32: I'm happy that these no longer have to live in `generic.cljc`!

src/sicmutils/polynomial.cljc
```diff
@@ -457,17 +457,16 @@
...
     ;; in polynomials over commutative rings should be factored out by an
     ;; expression analyzer before we get here. The result is a Polynomial object
     ;; representing the polynomial structure of the input over the unknowns.
     (let [expression-vars (sort v-compare (set/difference (x/variables-in expr) operators-known))
-          variables (zipmap expression-vars (a/new-variables this (count expression-vars)))]
-      (-> expr
-          (x/walk-expression variables operator-table)
-          (cont expression-vars))))
+          sym->var        (zipmap expression-vars (a/new-variables this (count expression-vars)))
+          expr'           (x/evaluate expr sym->var operator-table)]
```
sritchie, line 468: `x/evaluate` renaming is the big change here.

src/sicmutils/series/impl.cljc
```diff
@@ -450,7 +449,7 @@
 
 #_
 (= [0 1 1 1 1 1]
-   (take 6 (integral (iterate inc 1) 5)))
+   (take 6 (integral (iterate inc 1))))
```
sritchie, line 452: whoops, typo!

src/sicmutils/simplify.cljc
```diff
@@ -271,9 +271,6 @@
                         exp)]
     simplified-exp))
 
-(defmethod g/simplify [::x/numerical-expression] [a]
```
sritchie, line 274: I moved this to `sicmutils.abstract.number`, where we actually define these.

src/sicmutils/structure.cljc
```diff
@@ -497,5 +488,12 @@
 (defmethod g/cube [::structure] [a] (mul a (mul a a)))
 (defmethod g/simplify [::structure] [a] (->> a (mapr g/simplify) v/freeze))
 (defmethod g/transpose [::structure] [a] (opposite a (seq a)))
+
 (defmethod g/magnitude [::structure] [a]
-  (g/sqrt (reduce g/+ (map g/square a))))
```
sritchie, line 501: this wouldn't have worked with complex numbers before, since `g/square` performs a dot product, not an inner product. Now it does!

test/sicmutils/complex_test.cljc
```diff
@@ -195,19 +196,14 @@
       (is (near (g/div (g/sinh z) (g/cosh z))
                 (g/tanh z))))
 
-    ;; TODO enable these in the next PR, when we can properly force these
```
sritchie, line 198: woohoo, these work now that we don't push through the symbolic paths for defaults.

test/sicmutils/numbers_test.cljc
```diff
@@ -255,17 +255,13 @@
                        (g/cosh z))
                 (g/tanh z))))
 
-    ;; TODO enable these two in the next PR when this stops being a stack
```
sritchie, line 258: boom, defaults are eagerly evaluated so these work now.

test/sicmutils/numsymb_test.cljc
```diff
@@ -54,36 +55,36 @@
...
           "the symbolic operator is exact.")
-      (is (= -1 (g/cos 'pi)))
-      (is (near -1.0 (g/cos Math/PI)))
-      (is (= 1 (g/cos 'two-pi)))
-      (is (near 1.0 (g/cos (* 2 Math/PI))))
-      (is (= -1 (g/cos '-pi)))
-      (is (= 0 (g/cos 'pi-over-2)))))
+      (is (v/eq -1 (g/cos 'pi)))
+      (is (ish? -1 (g/cos Math/PI)))
+      (is (v/eq 1 (g/cos 'two-pi)))
```
sritchie, line 75: these return `(literal-number 1)`, which won't compare well in clj with the number on the left. `(= (g/cos 'pi) 1)` DOES work. better to use `v/eq` explicitly here.

test/sicmutils/polynomial_test.cljc
```diff
@@ -265,9 +265,9 @@
             (p/make 3 [[[0 0 1] 1]])] (a/new-variables poly-analyzer 3))))
 
   (testing "expr"
-    (let [exp1 (:expression (g/* (g/+ 1 'x) (g/+ -3 'x)))
```
sritchie, line 268: we no longer have a defrecord, so this keyword lookup needs to use a fn instead.

sritchie (commented):
@littleredcomputer , this is a big one but I think you're going to be happy!

src/sicmutils/function.cljc
sritchie: (comment
  "This comment comes from scmutils, function.scm, in the definition of
  `transpose-defining-relation`:

  T is a linear transformation T:V -> W
  the transpose of T, T^t:W* -> V*
  Forall a in V, g in W*,  g:W -> R
  (T^t(g))(a) = g(T(a)).")

from GJS. I ported the test that they include at the bottom of the page. Looks good!

src/sicmutils/ratio.cljc
```diff
@@ -217,6 +218,11 @@
      (defmethod g/gcd [Fraction Fraction] [^Fraction a ^Fraction b] (promote (.gcd a b)))
      (defmethod g/lcm [Fraction Fraction] [^Fraction a ^Fraction b] (promote (.lcm a b)))
      (defmethod g/expt [Fraction ::v/integral] [a b] (pow a b))
+     (defmethod g/sqrt [Fraction] [a]
```
sritchie, line 221: this was a trick GJS used in Scheme. Clojure does this trick too (via the numeric-tower library), but we need to add it here manually.

src/sicmutils/value.cljc
```diff
@@ -115,6 +140,27 @@
                        ::native-integral
                        ::floating-point)))
 
+  #?@(:clj
+      [java.lang.Double
+       (nullity? [x] (core-zero? x))
+       (unity? [x] (== 1 x))
+       (zero-like [_] 0.0)
```
sritchie, line 147: I added these explicitly so that we could get ACTUAL zero and one-likes, that kept their types, without muddying the `Number` implementation above.

src/sicmutils/value.cljc
```diff
@@ -136,7 +182,7 @@
 
   Symbol
   (nullity? [o] false)
-  (numerical? [_] false)
+  (numerical? [_] true)
```
sritchie, line 185: this makes more sense now, that symbols return `true` to this query; they are, without any wrapping required, "literal" numbers.

src/sicmutils/value.cljc
```diff
@@ -205,7 +251,7 @@
        number
        (-equiv [this other]
          (cond (core-number? other) (identical? this other)
-               (number? other)      (eq this other)
+               (numerical? other)   (eq this other)
```
sritchie, line 254: this lets Clojurescript compare numbers with numerical expressions like `(= 12 (literal-number 12))`.

sritchie:
@littleredcomputer , this has turned into a monster PR... there is a core that required some bigger changes, but beyond that I just went fairly ballistic adding tests and documentation.

I am MORE than happy to split this apart. But I think it stages us nicely for smaller changes down the road, so I'm glad it's all done.

CHANGELOG.md
```diff
@@ -6,45 +6,191 @@
...
 - fixed issues in `function.cljc` and `operator.cljc` where the Clojurescript
   `IFn` `-invoke` arguments shadowed either the `this` operator, or some
   parameter name in the deftype
   ([#169](https://github.com/littleredcomputer/sicmutils/pull/169))
 
+- `g/sqrt` now maintains precision with Clojurescript's rational numbers.
+  `(g/sqrt #sicm/ratio 9/4)` for example returns `#sicm/ratio 3/2`.
+  ([#168](https://github.com/littleredcomputer/sicmutils/pull/168))
+
+- `g/determinant` and `g/transpose` now act as identity for everything in the
```
littleredcomputer, line 56: Some books define the conjugate of a matrix as: transpose the matrix, then conjugate the entries elementwise, sometimes called the Hermitian conjugate. Scmutils agrees with you that the operation is identity. I am very surprised GJS did not take this bait. Tempted?

src/sicmutils/numsymb.cljc
```diff
@@ -22,59 +22,67 @@
...
 (def sqrt? (is-expression? 'sqrt))
 (def expt? (is-expression? 'expt))
-(def quotient? (is-expression? (symbol "/")))
+(def quotient? (is-expression? '/))
 (def arctan? (is-expression? 'atan))
 (def operator first)
 (def operands rest)
 
-;; BEGIN
+(defn- delegator
```
littleredcomputer, line 46: I might name this `with-exactness-preserved`

src/sicmutils/structure.cljc
littleredcomputer: I thought long and hard about it at the time: is GJS's definition of `square` principled? I concluded that he very much wants to write (* 1/2 m (square v)) with this generalizing to structured v. Which makes me wonder whether the actual definition of square we should advertise is (* (transpose v) v)

test/sicmutils/function_test.cljc
```diff
@@ -139,6 +168,73 @@
...
+                 (* a↑0 g_1 (((partial 0) T↑1) (up x y)))
+                 (* a↑0 g_2 (((partial 0) T↑2) (up x y)))
+                 (* a↑1 g_0 (((partial 1) T↑0) (up x y)))
+                 (* a↑1 g_1 (((partial 1) T↑1) (up x y)))
+                 (* a↑1 g_2 (((partial 1) T↑2) (up x y))))
+             (g/simplify
+              (((g/transpose (DTf s)) g) a)))
+          "A wild test of transpose from scmutils function.scm that I don't
+    understand!"))))
+
```
littleredcomputer, line 237: I was thinking about the GJS definition of transpose for function and now think it might be clearer if we wrote 

    T  :  V → W
    g  :  W → R
    T' : (W → R) → (V → R)
    T' :    g    ↦  g ∘ T

(... because, how else can you do it? )

littleredcomputer (approved):

src/sicmutils/structure.cljc
sritchie: That was my impression too, that this is a glitch in the matrix for that purpose.

Your definition is clearer and matches the scmutils behavior:

``scheme
2 error> (square (abs (up 1+2i 2 3)))
#| 10.+4.000000000000001i |#

2 error> (square (magnitude (up 1+2i 2 3)))
#| 17.999999999999996 |#

2 error> (square (up 1+2i 2 3))
#| 10+4i |#
```

File: CHANGELOG.md [hunk 34]
```diff
@@ -6,45 +6,191 @@
...
+- If you want to compare literal numbers and an expression like
+  `(an/literal-number 12)`, use `v/eq`. In Clojurescript, this will work with
+  the built in `=` as well, since equality is implemented with a protocol that
+  we can extend. For example:
+
+```clojure
+(v/eq 12 (literal-number 12))
+;;=> true
+
+(= 12 (literal-number 12))
```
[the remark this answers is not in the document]
sritchie: yes! generic comparisons in clj will fix this. It also works just fine in Clojurescript because it's protocols from the bottom up and we can hook in generic implementations at the language level.

src/sicmutils/expression.cljc
```diff
@@ -18,70 +18,242 @@
...
+  Object
+  (toString [_] (str expression))
+  #?(:clj
+     (equals [a b]
+             (if (instance? Literal b)
+               (let [b ^Literal b]
+                 (and (= type (.-type b))
+                      (= expression (.-expression b))
+                      (= meta (.-meta b))))
+               (v/eq a b))))
```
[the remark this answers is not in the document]
sritchie: hey, yes, sorry about this... `v/eq` is our generic "equals" operator that I had to add a while back, to get native Clojurescript types to `=` properly with

- rational numbers
- `js/BigInt`
- `goog.math.{Integer, Long}`

without this override, we get

image

In this exact spot, this lets me write 

```
sicmutils.env>    (= (literal-number 12) 12)
true
```

GJS actually has an implementation in scmutils that will attempt to simplify down to a number inside his `zero?`, `=` and `one?` methods. Note his comments, and his "sigh." below...

```
;;; Conservative tests...  These tests will return TRUE only if the
;;; default simplifier can prove that the answer is TRUE.  This is a
;;; bad idea to make generic, because it puts a simplification burden
;;; on lots of parts of the system. -- GJS

(define (an:= x y)
  (an:zero? (g:- x y)))

(define (an:zero? x)
  (let* ((ex (expression x))
	 (evars
	  (list-difference (variables-in ex) symbolic-operators))
	 (val
	  (ignore-errors
	   (lambda ()
	     (apply (eval `(lambda ,evars ,ex) symbolic-environment)
		    (map (lambda (x) (random 10000)) evars))))))
    (if (and (not (condition? val)) (exact-zero? val))
	(exact-zero? (simplify ex))
	#f)))

(define (an:one? x)
  (exact-one? (g:simplify x)))

;;; Sigh.
;;; (assign-operation '=   an:=  abstract-number? abstract-number?)
;;; (assign-operation 'zero?  an:zero?        abstract-number?)
;;; (assign-operation 'one?   an:one?         abstract-number?)
```

test/sicmutils/function_test.cljc
sritchie: That makes more sense. When he says `T^t : W* -> V*`, does `W*` mean that, some fn from `W` to whatever?

sritchie: i updated the comment here, and I see how to run this example through now, no problem. I need to stare at it for a while to think about why he chose this particular example.

src/sicmutils/numsymb.cljc
sritchie: will do! I will add this as a final PR once I merge the chain.

CHANGELOG.md
sritchie: YES! I'll add this in a follow-up.

sritchie:
Thanks again for wading through this monster...

test/sicmutils/function_test.cljc
littleredcomputer: `W*` (the star should be superscript) means the "dual vector space" to W, which is the space of linear functionals on W. "functional" is jargon for "a function to the ground ring" (here assumed to be the reals R). In Euclidean space, it turns out that every linear functional can be written as a row vector. So that's why I write that T' (by which I meant T^t) takes W->R to V->R, i.e., it is the dual of the map T, and takes arguments from the dual vector spaces _in the opposite direction_, and I left the star-notation behind (but it should be emphasized that the maps are linear). D takes a function, and a point, and gives you back a linear functional: the row vector which can combine with column vectors representing directions to tell you how much the function will increase in that direction for a short distance. This is why D exchanges up  and down.

PR #169, opened 2020-11-30
matrix and structure literals, new exp and log fns

sritchie:
Continuing the expansion of our stable of generics. This PR adds:

- `literal-matrix` to generate symbolic matrices
- `literal`, `literal-up` and `literal-down` to generate symbolic structures

These are both exposed in `sicmutils.env`.

### New generics:

- `g/exp2`, `g/exp10` for exponents with base 2 and 10
- `g/log2`, for base 2 logarithms
- `g/log10` for base 10 logs

also, `g/lcm` is now exposed in `sicmutils.env`

### Misc

I also fixed a bug in `function.cljc` and `operator.cljc`, with `-invoke` arguments shadowing in the higher arities!


littleredcomputer (approved):
brilliant. (thanks for the up arrow. It turns out to be used _a lot_ in FDG)

PR #170, opened 2020-11-30
Documentation push, prepping for cljdoc

sritchie:
This PR:

- adds a table of contents for a solid https://cljdoc.org/ build
- imports refman.txt! I've started converting this to Clojure, but it will take a while. I added a note that we're in progress.
- creates placeholder files for all of the docs we should write.


sritchie:
@littleredcomputer , I am going to recommend that we figure out together if the table of contents looks good, and if so then go ahead and merge this, placeholders and all. It won't get published until we can cut a release, and this will give me a structure to start filling in with writing.

wdyt?

littleredcomputer:
that sounds fine to me

littleredcomputer (commented):
This breakdown of the structure is really cool.

PR #171, opened 2020-12-01
move literal-function => abstract, switch fn inheritance order

sritchie:
This PR 

- switches the inheritance direction of `::v/function` and `::sicmutils.function`
- moves all of the `literal-function`machinery and tests into their own namespace, `sicmutils.abstract.function`
- keeps `sicmutils.function` quite clean. This gets the library closer to a "built from Clojure up" style.

::v/function` is now the main thing, and `::sicmutils.function/function` extends from this.


src/sicmutils/abstract/function.cljc
```diff
@@ -0,0 +1,274 @@
...
+;; This software is distributed in the hope that it will be useful, but
+;; WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;; General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this code; if not, see <http://www.gnu.org/licenses/>.
+;;
+
+(ns sicmutils.abstract.function
```
sritchie, line 20: just a move, nothing new.

src/sicmutils/abstract/number.cljc
```diff
@@ -110,6 +110,13 @@
     (x/expression-of expr)
     expr))
 
+(defn- defunary [generic-op op-sym]
```
sritchie, line 113: I moved this up to keep unary => binary order...

src/sicmutils/calculus/derivative.cljc
```diff
@@ -431,14 +432,11 @@
           ((d f) (first xs))
           ((d #(apply f %)) (matrix/seq-> xs)))))))
 
-(derive ::x/numeric ::codiff)
```
sritchie, line 434: I had introduced this and forgotten that `::scalar` was all we really need, no need now for a new `::codiff`.

src/sicmutils/calculus/derivative.cljc
```diff
@@ -493,11 +491,14 @@
 (defmethod g/partial-derivative [::matrix/matrix v/seqtype] [f selectors]
   (multivariate-derivative f selectors))
 
+(def derivative-symbol 'D)
```
sritchie, line 494: I moved this in here so it lives in this one spot.

src/sicmutils/matrix.cljc
```diff
@@ -31,6 +32,8 @@
 
 (declare generate)
 
+(derive ::matrix ::f/cofunction)
```
sritchie, line 35: @littleredcomputer , curious what you think here. I realized that if we were building this up from scratch, we wouldn't have matrices at the point where we added functions... so I moved this derivation over to `matrix`, so that `function.cljc` could stay cleaner.

What do you think?

If someone ever wants to use a matrix, they have to require this `ns` anyway, which will trigger the `derive` call. And if they don't, no sweat.

src/sicmutils/series.cljc
```diff
@@ -563,6 +563,7 @@
 
 (derive ::x/numeric ::coseries)
 (derive ::v/number ::coseries)
+(derive ::v/function ::coseries)
```
sritchie, line 566: same thought here as in matrix; series should register themselves.

src/sicmutils/structure.cljc
```diff
@@ -41,10 +42,14 @@
 (def ^:private opposite-orientation
   {::up ::down ::down ::up})
 
+
 (derive ::up ::structure)
 (derive ::down ::structure)
 (derive PersistentVector ::up)
 
+;; Structures can interact with functions.
+(derive ::structure ::f/cofunction)
```
sritchie, line 51: same thought here.

test/sicmutils/abstract/function_test.cljc
```diff
@@ -0,0 +1,240 @@
...
+;; This software is distributed in the hope that it will be useful, but
+;; WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;; General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this code; if not, see <http://www.gnu.org/licenses/>.
+;;
+
+(ns sicmutils.abstract.function-test
```
sritchie, line 20: these are all the tests that explicitly use `literal-function`.

test/sicmutils/function_test.cljc
```diff
@@ -18,66 +18,20 @@
...
-  (:refer-clojure :exclude [partial])
   (:require [clojure.test :refer [is deftest testing use-fixtures]]
-            [clojure.test.check.generators :as gen]
-            [com.gfredericks.test.chuck.clojure-test :refer [checking]
-             #?@(:cljs [:include-macros true])]
-            [sicmutils.calculus.derivative :refer [D partial]]
+            [sicmutils.calculus.derivative :refer [D]]
             [sicmutils.generic :as g]
-            [sicmutils.generators :as sg]
-            [sicmutils.matrix :as m]
```
sritchie, line 29: The only tests left here are the ones that did NOT use literal-function. Now we can add more, only about Clojure functions.

littleredcomputer (approved):
LGTM

PR #172, opened 2020-12-03
Remove explicit literal up and down tuple creation in internal fns

sritchie:
This PR removes the few spots where we were explicitly constructing explicit literal-up and literal-down tuples, via `literal-function`.


src/sicmutils/calculus/form_field.cljc
```diff
@@ -121,13 +121,10 @@
...
-        domain (apply s/up (repeat n 0))
-        range 0
-        components (s/generate n ::s/down #(af/literal-function
-                                            (symbol (str name \_ %))
-                                            domain
-                                            range))]
-    (components->oneform-field components coordinate-system name)))
+        domain (apply s/up   (repeat n 0))
+        range  (apply s/down (repeat n 0))]
+    (-> (af/literal-function name domain range)
```
littleredcomputer, line 126: I guess GJS would use `UP*` here. When I look at the usage here, I wonder what the notation really ought to be for this.

littleredcomputer (approved):

src/sicmutils/calculus/form_field.cljc
sritchie: I still don't know this well enough to suggest an alternative, so let's follow your lead here if we should change this.

PR #176, opened 2020-12-03
Beef up def-generic-function with metadata support

sritchie:
We already allow two metadata keys... why not open up the field of play?

We don't use this yet, but we could if we want to add more metadata about what sorts of things each operation supports. Right now the set is totally open.


src/sicmutils/generic.cljc
```diff
@@ -55,29 +68,31 @@
...
     `(do
        (defmulti ~f ~docstring v/argument-kind ~@options)
        (defmethod ~f [~kwd-klass] [k#]
-         ({:arity ~arity :name '~name} k#)))))
+         (~attr k#)))))
 
 ;; Numeric functions.
-(def-generic-function add 2    {:name +})
-(def-generic-function negate 1 {:name -})
+(def-generic-function add 2    {:name '+})
```
sritchie, line 80: it's better anyway that this requires quoting vs the previous method.

littleredcomputer (approved):

PR #178, opened 2020-12-03
Add more generic operations: trace, dimension

sritchie:
This PR adds:

- `g/trace` for square matrices and square structures (acts as identity for all scalars)
- `g/dimension` (always 1 for scalars) for scalars, square, row and column matrices, and structures
- `m/nth-col` and `m/diagonal`, `m/column?`, `m/row?` in matrix.cljc

we also have three new `kind` keywords for matrices. Square matrices return`::m/square-matrix`, and columns and rows return `::m/column-matrix` and `::row-matrix` respectively. These all derive from `::m/matrix`. This makes it easier to register methods or test specifically for these cases. We've also added `m/column?` and `m/row?` predicates to check for these cases.


sritchie:
Okay @littleredcomputer, this is ready and is the last one in this big push!

littleredcomputer (approved):

src/sicmutils/matrix.cljc
```diff
@@ -491,8 +533,6 @@
     (when-not (= r c) (u/illegal "not square"))
     (determinant (g/- (g/* x (I r)) m))))
```
littleredcomputer, line 534: Before I forget: I was looking at this and thinking, "it would be cool if `(characteristic-polynomial M)` returned a function of `x` that was polynomial in x, without having to name it. Of course, that is easily done, by just having the function return `(fn [x] ...)`, but then it does a (potentially vast) amount of computation every time you call it. If only there were a way to record that computation in a way that could be played back on `'x` or `3`.  I think that is something we should think about; it would have application to our numerics and to differentiation itself.

sritchie: noted, writing this down to noodle over.

PR #179, opened 2020-12-05
restrict numsymb access to `symbolic-operator` fn

sritchie:
This allows us to make the interface fully closed, and properly switch between the "constructor simplification" world and the current implementations when we add those.


src/sicmutils/polynomial.cljc
```diff
@@ -474,17 +474,20 @@
...
-           (sort-by exponents #(monomial-order %2 %1))
-           (map (fn [[xs c]]
-                  (->> (map (fn [exponent var]
-                              (sym/expt var exponent))
-                            xs vars)
-                       (reduce sym/mul 1)
-                       (sym/mul c))))
-           (reduce sym/add 0))
-      p))
+    (let [*    (sym/symbolic-operator '*)
```
sritchie, line 477: @littleredcomputer , this is the flavor of change I pushed through. Let me know what you think here.

src/sicmutils/simplify.cljc
```diff
@@ -18,19 +18,16 @@
 ;
 
 (ns sicmutils.simplify
-  (:require [clojure.walk :refer [postwalk]]
```
sritchie, line 21: removed extra imports here.

src/sicmutils/polynomial.cljc
littleredcomputer: It seems to work out to the same thing, more or less? The definition of symbolic-operator knows the definition of sym/add ...

littleredcomputer (approved):

src/sicmutils/polynomial.cljc
sritchie: Yes, for now it does. I want to add (for completeness, oh how I'm regretting thinking this way) the ability to toggle the more-advanced constructor simplifications from `numsymb.scm` on and off with a dynamic variable. This one step of indirection makes that possible.

PR #180, opened 2020-12-05
Rename nulllity? unity? to zero?, one?

sritchie:
@littleredcomputer, happy to ditch this if it's controversial - but it looks like a straightforward change, and, imo, makes generic implementations read more naturally.

Let me know what you think!


littleredcomputer (approved):
I can't remember why I didn't do this but there was a (perhaps ephemeral) reason. Go for it!

PR #184, opened 2020-12-06
Move function and symbol `Value` implementations => their respective namespaces

sritchie:
This PR:

- moves the `Symbol` value implementation into `sicmutils.abstract.number`
- adds explicit `Value` implementations for the various function types, and moves these (and the `arity` code) to `sicmutils.function`
- more generic function implementations for `::v/function`
- lots more generative tests for functions, numbers
- adds clearer `Approximate` implementations for our various types, so all numbers now compare appropriately to complex and each other with `ish?`

Plus a couple small fixes.


src/sicmutils/abstract/number.cljc
```diff
@@ -28,6 +28,17 @@
   #?(:clj
      (:import (clojure.lang Symbol))))
 
+(extend-protocol v/Value
+  Symbol
+  (zero? [o] false)
+  (one? [_] false)
+  (zero-like [_] 0)
+  (one-like [_] 1)
+  (exact? [sym] (:exact? (meta sym) true))
```
sritchie, line 37: before, we had symbols as inexact... but in fact, in scmutils (and this makes sense now), a symbol is treated as an exact, literal number.

In the next PR I'll add metadata support to abstract expressions, which will allow BOTH to work with this logic.

src/sicmutils/complex.cljc
```diff
@@ -35,9 +35,10 @@
 (defn complex
   "Construct a complex number from real, or real and imaginary, components."
   ([re]
-   (Complex. re))
```
sritchie, line 38: this bug was exposed by the new tests in `number.cljc`... passing in a `goog.math.Long`, for example, broke this method.

src/sicmutils/env.cljc
```diff
@@ -302,7 +299,10 @@
   state-advancer]
  [sicmutils.numerical.quadrature definite-integral]
  [sicmutils.numerical.elliptic elliptic-f]
- [sicmutils.numerical.minimize minimize multidimensional-minimize])
+ [sicmutils.numerical.minimize minimize multidimensional-minimize]
+ [sicmutils.value exact? zero? one?
```
sritchie, line 303: more of these are useful, so let's import the batch!

src/sicmutils/function.cljc
```diff
@@ -30,16 +35,225 @@
...
   like to know the arity of our functions, so that we can calculate
   their derivatives with structure, etc. The arity of a composition is
   simply the arity of its rightmost (that is, first to be applied)
   function term."
   [& fns]
-  (let [a (v/arity (last fns))]
+  (let [a (arity (last fns))]
     (with-meta (apply comp fns) {:arity a})))
 
+(defn- zero-like [f]
```
sritchie, line 50: the main change here is that I've pulled the `Value` implementation into this file, and was able to bring the `arity` functions along for the ride. Now, `value` is closer to a very basic namespace that just starts off the library by defining the protocols we care about. The function algebra all lives in `sicmutils.function`, and `sicmutils.abstract.function` extends it to abstract.

src/sicmutils/function.cljc
```diff
@@ -30,16 +35,225 @@
...
+  (zero? [_] false)
+  (zero-like [f] (zero-like f))
+  (one? [_] false)
+  (one-like [f] (one-like f))
+  (exact? [f] (compose v/exact? f))
+  (numerical? [_] false)
+  (freeze [f] (get @v/object-name-map f f))
+  (kind [_] ::v/function)
+
+  Var
```
sritchie, line 88: oh, we get `var` support too!

src/sicmutils/function.cljc
```diff
@@ -126,12 +340,19 @@
 (defbinary g/add g/+)
 (defbinary g/sub g/-)
 (defbinary g/mul g/*)
+(defunary g/invert)
```
sritchie, line 343: I realized these were missing in the course of my moves, and added tests for these too.

src/sicmutils/numbers.cljc
```diff
@@ -218,7 +218,21 @@
          (op (js/Number a) b))
 
        (defmethod op [::v/floating-point js/BigInt] [a b]
-         (op a (js/Number b))))
+         (op a (js/Number b)))
+
+       ;; BigInt can't handle these operations natively, so we override with a
+       ;; downcast to number for now.
+       (doseq [op [g/cos g/sin g/tan
```
sritchie, line 225: This actually showed up in #168; I just added it before the PR chain left the station and haven't rolled that change forward. I'll resolve when I merge.

src/sicmutils/numerical/derivative.cljc
```diff
@@ -486,7 +486,7 @@
   "Fills in default values required by `D-numeric`. Any option not used by
   `D-numeric` gets passed on to `us/seq-limit`."
   [m]
-  (let [defaults {:tolerance (Math/sqrt v/machine-epsilon)
+  (let [defaults {:tolerance v/sqrt-machine-epsilon
```
sritchie, line 489: we added this in a previous PR, might as well use it.

src/sicmutils/rational_function.cljc
```diff
@@ -35,6 +35,10 @@
   v/Value
   (zero? [_] (v/zero? u))
   (one? [_] (and (v/one? u) (v/one? v)))
+  (zero-like [_]
```
sritchie, line 38: I took a look at the scmutils code, and it looks like they are returning polynomials in these cases... I'm just fleshing out `Value` support, not changing behavior like that for now.

src/sicmutils/value.cljc
```diff
@@ -140,6 +136,16 @@
                        ::native-integral
                        ::floating-point)))
 
+  #?(:clj Boolean :cljs boolean)
```
sritchie, line 139: this came up because `g/negative?` returns a boolean, and we need `(g/one-like g/negative?)` to work, and return a one-like in the proper range.

I think this is preferable to throwing. wdyt @littleredcomputer ?

test/sicmutils/abstract/function_test.cljc
```diff
@@ -33,11 +33,39 @@
             [sicmutils.structure :as s :refer [literal-up
                                                literal-down
                                                up down]]
-            [sicmutils.simplify :as ss :refer [hermetic-simplify-fixture]]))
+            [sicmutils.simplify :refer [hermetic-simplify-fixture]]))
 
 (use-fixtures :each hermetic-simplify-fixture)
 
-(def ^:private near (v/within 1.0e-6))
+(deftest value-protocol-tests
```
sritchie, line 40: new tests for the value implementation of abstract function.

test/sicmutils/abstract/function_test.cljc
```diff
@@ -94,6 +122,35 @@
               (is (= (g/magnitude (f z)) ((g/magnitude f) z)))
               (is (= (g/angle (f z))     ((g/angle f) z))))))
 
+(deftest trig-tests
```
sritchie, line 125: these aren't explicitly literal function tests, but they're function tests that have to do with symbolic inputs, so they seemed to belong here.

test/sicmutils/abstract/number_test.cljc
```diff
@@ -96,11 +100,6 @@
                       (+ x 0.5)
                       x))))))
 
-(def double-or-int
```
sritchie, line 99: another change from a previous PR, mirrored here.

test/sicmutils/calculus/derivative_test.cljc
```diff
@@ -40,7 +40,7 @@
   #?(:clj
      (:import [sicmutils.calculus.derivative Differential])))
 
-(use-fixtures :once hermetic-simplify-fixture)
+(use-fixtures :each hermetic-simplify-fixture)
```
sritchie, line 43: we have some issues with the simplifier not sorting products and sums; this makes the whole thing more repeatable by preventing cache collisions between different tests.

Unfortunately it also changes up some of the test results below (just commutative changes, nothing bad).

test/sicmutils/calculus/derivative_test.cljc
```diff
@@ -590,6 +590,22 @@
              (g/simplify ((as-matrix (D C-general)) s)))))))
 
 (deftest taylor-moved-from-series
+  (let [simp4 (fn [x] (g/simplify (take 4 x)))
```
sritchie, line 593: oh, I should also note that I moved some tests OUT of function, so that I could keep sicmutils.function-test containing only tests of function behavior.

I'm trying to sort the library into clear, logical layers, so at the "time of implementation" of function.cljc, series, operators and derivatives don't exist yet.

So those tests moved out to their respective spots!

test/sicmutils/complex_test.cljc
```diff
@@ -50,7 +50,7 @@
 (deftest complex-laws
   ;; Complex numbers form a field. We use a custom comparator to control some
   ;; precision loss.
-  (binding [sg/*complex-tolerance* 1e-3]
+  (with-comparator (v/within 1e-3)
```
sritchie, line 53: I fixed things up in `generators.cljc` so that we don't need this special handling for complex numbers.

test/sicmutils/function_test.cljc
```diff
@@ -18,22 +18,49 @@
...
-
-(use-fixtures :once hermetic-simplify-fixture)
-
-(def ^:private near (v/within 1.0e-6))
+            [sicmutils.numbers]
+            [sicmutils.value :as v]))
 
 (deftest value-protocol-tests
-  (testing "freeze"
+  (testing "v/zero? returns false for fns"
```
sritchie, line 34: these tests are all new.

test/sicmutils/function_test.cljc
```diff
@@ -46,97 +73,164 @@
...
-    (let [f (f/compose g/cos g/acos)]
-      (is (near 0.5 (f 0.5)))
-
-      (testing "outside real range"
-        (is (near 5 (g/magnitude (f -5)))
-            "This kicks out a complex number, which doesn't yet compare
-          immediately with reals."))))
-
-  (testing "tan/atan"
+  (with-comparator (v/within 1e-8)
```
sritchie, line 160: I converted a bunch of these tests to generate tests that work on numbers (instead of symbols), and moved the symbol tests to abstract.function-test.

test/sicmutils/function_test.cljc
```diff
@@ -157,14 +251,6 @@
     returns a fn that applies them in reverse order. Like a curried andThen (the
     reverse of compose)"))))
 
-(deftest string-form-test
```
sritchie, line 160: this moved to the simplifier tests.

test/sicmutils/function_test.cljc
```diff
@@ -190,14 +276,50 @@
...
 
-    (testing "cross-product"
-      (let [deferred (g/cross-product #(g/* 2 %)
-                                      #(g/+ (s/up 4 3 1) %))
-            v (s/up 1 2 3)]
-        (is (= (g/cross-product (g/* 2 v)
-                                (g/+ (s/up 4 3 1) v))
-               (deferred v))
-            "Slightly tougher since this works with structures")))
+    (checking "invert" 100 [n sg/real]
```
sritchie, line 279: these are tests of all the new generic methods implemented for fns.

test/sicmutils/function_test.cljc
```diff
@@ -227,43 +349,9 @@
...
-        (is (= [:at-least 0] (v/arity mul)))
-        (is (= [:at-least 0] (v/arity add+mul)))
+        (is (= [:at-least 0] (f/arity add)))
+        (is (= [:at-least 0] (f/arity mul)))
+        (is (= [:at-least 0] (f/arity add+mul)))
         (is (= 33 (add+mul 2 3 4)))
         (is (= -15 (add-mul 2 3 4)))
         (is (= 15 (mul-add 2 3 4)))))))
-
-(deftest operators
```
sritchie, line 237: moved to operator, since "operator" doesn't "exist yet", as far as the function implementation is concerned.

test/sicmutils/generators.cljc
```diff
@@ -61,6 +63,9 @@
                long
                integer]))
 
+(def real
```
sritchie, line 66: this came in from a previous PR.

test/sicmutils/generators.cljc
```diff
@@ -81,15 +86,57 @@
               d)]
       (r/rationalize n d))))
 
-(def ^:dynamic *complex-tolerance* 1e-12)
+(defn square-matrix
```
sritchie, line 89: I moved this in from `matrix_test.cljc` so we could start consolidating our generators.

test/sicmutils/generators.cljc
```diff
@@ -81,15 +86,57 @@
...
+  ([n] (square-matrix n ratio))
+  ([n entry-gen]
+   (gen/fmap #(apply m/by-rows %)
+             (gen/vector (gen/vector entry-gen n) n))))
+
+(defn- int=? [this that]
```
sritchie, line 95: this and the following protocol extension allow all of our real types to compare approximately with complex.

test/sicmutils/numbers_test.cljc
```diff
@@ -237,62 +231,88 @@
     (is (c/complex? (g/asin 2)))
     (is (c/complex? (g/acos 2))))
 
-  (testing "sin"
```
sritchie, line 240: woohoo, all of these are now generative tests.

src/sicmutils/value.cljc
littleredcomputer: isn't that the very definition of a bike shed. I think the answer is not obvious (I could argue that the correct answer is "the identity function"), while the stakes are vanishingly low

test/sicmutils/generators.cljc
```diff
@@ -81,15 +86,57 @@
...
+  (cond (c/complex? that) (and (si/*comparator* 0.0 (g/imag-part that))
+                               (si/*comparator*
+                                (u/double this)
+                                (g/real-part that)))
+        (v/real? that)    (si/*comparator*
+                           (u/double this)
+                           (u/double that))
+        :else             (= this that)))
 
 (extend-protocol si/Approximate
```
littleredcomputer, line 105: Ya learn something new every day!

test/sicmutils/generators.cljc
littleredcomputer: what does int= mean? Is one of these guaranteed to be complex?

littleredcomputer (commented):
cool, but that seems like a lot of kerfuffle for complex

src/sicmutils/value.cljc
sritchie: yes, or `true`, depending on the algebra.

The only plea I can enter here for the "why bikeshed?" charge is that SOME decision was forced on me by generative testing!

test/sicmutils/generators.cljc
sritchie: The `same` library has been great... this idea comes up in every damned library. Why can't a language give it to us out of the box??

PR #185, opened 2020-12-07
Add metadata implementation to expressions, properly freeze modint

sritchie:
This PR:

- specializes g/exp and the trig operations to square matrices (forgot to do this previously!)
- adds a proper metadata implementation to `Literal`! We can use this now instead of my hand-rolled thing I was starting to do. Better to stick to Clojure's great interfaces.
- adds a `v/freeze` implementation to modint, another small thing I noticed.


src/sicmutils/expression.cljc
```diff
@@ -66,17 +68,28 @@
...
 
-  #?@(:cljs
-      [IEquiv
+  #?@(:clj
+      [IObj
+       (meta [_] m)
+       (withMeta [_ meta] (Literal. type expression meta))]
+
+      :cljs
+      [IMeta
```
littleredcomputer, line 80: Is there a circumstance where you could have Meta but not WithMeta?

src/sicmutils/matrix.cljc
```diff
@@ -542,19 +542,19 @@
...
-(defmethod g/sec [::matrix] [m] (series/sec-series m))
-(defmethod g/acos [::matrix] [m] (series/acos-series m))
-(defmethod g/asin [::matrix] [m] (series/asin-series m))
-(defmethod g/atan [::matrix] [m] (series/atan-series m))
-(defmethod g/cosh [::matrix] [m] (series/cosh-series m))
-(defmethod g/sinh [::matrix] [m] (series/sinh-series m))
-(defmethod g/tanh [::matrix] [m] (series/tanh-series m))
-(defmethod g/asinh [::matrix] [m] (series/asinh-series m))
-(defmethod g/atanh [::matrix] [m] (series/atanh-series m))
+(defmethod g/exp [::square-matrix] [m] (series/exp-series m))
```
littleredcomputer, line 545: I love ::square-matrix, that's a good idea.

littleredcomputer (approved):

src/sicmutils/expression.cljc
sritchie: scrolling the cljs source, it looks like you would do this for a type you don't control; the `withMeta` implementation would have to go onto a wrapper type, `MetaFn` in the case of functions with cljs.

`meta` on a bare function returns `nil`.

littleredcomputer: Oh, I was confused. I thought IWithMeta was a name you had chosen.

Are we going to maintain a difference between metadata as we see it and the native implementation?

sritchie: Nope, this actually brings us back to how native works!

`defrecord` gets support for metadata out of the box, so that's why it worked before if it did.

But deftypes don't, so you have to manually carry around the map and implement these protocols.

PR #189, opened 2020-12-07
Add `make-rectangular`, `make-polar` for functions, numbers, symbolic types

sritchie:
This PR:

- adds a `sicmutils.complex/I` alias for `i`
- Adds the following two generic operators:
  - `g/make-rectangular`, (build a complex number from real and imaginary parts)
  - `g/make-polar` (build a complex number from radius and angle)

these work with real numbers, symbolic numbers, functions and any combination of these. They match the scmutils behavior in all cases.

It WOULD be nice to further extend these to handle up, down and matrices, for constructing complex structures/matrices where the shapes match.


src/sicmutils/numbers.cljc
```diff
@@ -58,7 +58,13 @@
 ;; ## Complex Operations
 (defmethod g/real-part [::v/real] [a] a)
 (defmethod g/imag-part [::v/real] [a] 0)
-(defmethod g/angle [::v/real] [a] (v/zero-like a))
+
+(let [neg-pi (core-minus Math/PI)]
```
sritchie, line 62: I know I had this in a previous PR and didn't push it forward. I'll deal with the merge when I get there...

test/sicmutils/abstract/number_test.cljc
```diff
@@ -499,6 +515,52 @@
...
+            (is (v/= (g/+ sym (g/* c/I n))
+                     (g/make-rectangular sym n))))
+
+  (checking "make-polar" 100 [r     gen/symbol
+                              theta gen/symbol]
+            (is (= (g/* r (g/+ (g/cos theta)
+                               (g/* c/I (g/sin theta))))
+                   (g/make-polar r theta))))
+
+  (checking "make-polar with numbers" 100
```
littleredcomputer, line 537: it is so nice to have tests like these

littleredcomputer (commented):
awesome

test/sicmutils/abstract/number_test.cljc
sritchie: This project has really gotten me used to leaning ONLY on tests like this and being less comfortable with only bare unit tests. A win!

PR #196, opened 2020-12-10
Add transpose laws, more tests

sritchie:
This PR pulls in laws for `transpose-outer`, `g/transpose` and `s:transpose` from the `scmutils` implementation. This makes me feel even better about our change in the previous PR.

### DIFF
--- a/test/sicmutils/matrix_test.cljc
+++ b/test/sicmutils/matrix_test.cljc
@@ -30,6 +30,7 @@
             [sicmutils.matrix :as m]
             [sicmutils.numsymb]
             [sicmutils.structure :as s]
+            [sicmutils.structure-test :refer [<l|:inner:|r>]]
             [sicmutils.value :as v]))
 
 (deftest value-protocol-tests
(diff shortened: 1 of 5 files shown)


sritchie:
@littleredcomputer , I'm going to merge (since there's no new stuff, just tests), but would love eyes when you get a moment.

test/sicmutils/matrix_test.cljc
```diff
@@ -226,6 +227,26 @@
...
 (deftest structure
+  (checking "(s:transpose <l|, inner, |r>)==(s/transpose-outer inner)"
+            100 [[l inner r] (gen/let [rows (gen/choose 1 5)
+                                       cols (gen/choose 1 5)]
+                               (<l|:inner:|r> rows cols))]
+            (is (v/zero?
+                 (g/- (m/s:transpose l inner r)
+                      (s/transpose-outer inner)))))
+
+  ;; this binding is required because `s:transpose` is meant to return a scalar;
```
littleredcomputer, line 238: I'm not totally sure I follow this ... do you think it's good math or are we verifying compatibility with GJS here?

sritchie: yeah, I don't follow this either!! This is to verify compatibility with GJS, until I figure out more about what the heck is happening here.

PR #197, opened 2020-12-10
new functions + matrix operations, gearing up for release

sritchie:
This PR gets us up to ~parity with the functions exposed in `refman.txt`. The changelog entries are:

- `matrix.cljc` has a new `by-cols` (analogous to `m/by-rows`), and `row` to
  generate a row matrix (analagous to `column`).
  [#197](https://github.com/littleredcomputer/sicmutils/pull/197) Also in
  `matrix.cljc`:

  - `num-rows`, `num-cols` access the row or column number without inspecting
    the deftype variables directly
  - `with-substituted-row`, for swapping out a single row in a matrix
  - `submatrix` generates a submatrix from low and high row and cols
  - `matrix-some` renamed to `some`: make sure to use a namespace prefix to
    avoid clashing with `clojure.core/some`.
  - new-matrix constructor `by-cols` (analogous to `by-rows`, takes a sequence
    of columns)
  - `row` constructor takes a sequence of values and returns a row matrix.
  - `by-rows*`, `by-cols*`, `row*` and `column*` are non-variadic versions of
    those functions. If you already have a sequence of rows, columns or
    elements, prefer these.
  - `up->row-matrix` => `down->row-matrix` and `row-matrix->up` =>
    `row-matrix->down`. A row is analogous to a `down`, so we make a change to
    reflect this.
  - `g/cross-product` between two `down` structures now returns a `down`.
  - `make-zero` generates a zero-valued matrix of the supplied dimensions.
  - `make-diagonal` generates a diagonal matrix containing the values of the
    supplied sequence.
  - `identity?` now returns true for identity matrices, false
    otherwise. `one?` returns false for identity matrices.
  - `one-like` and `identity-like` do the right thing for their contained types.

- `sicmutils.structure/up` and `sicmutils.structure/down` now have analogous
  `s/up*` and `s/down*` functions. These behave identically, but are
  non-variadic. If you already have a sequence you'd like to transform, prefer
  these ([#197](https://github.com/littleredcomputer/sicmutils/pull/197)).

- `sicmutils.value/kind-predicate` takes some item and returns a predicate that
  returns true if its argument has the same type (or inherits from it)
  ([#197](https://github.com/littleredcomputer/sicmutils/pull/197)).

- `sicmutils.function/arg-shift` and `sicmutils.function/arg-scale` take
  functions and return new functions that shift and scale their arguments
  (respectively) by the originally supplied shifts
  ([#197](https://github.com/littleredcomputer/sicmutils/pull/197)).

- `sicmutils.generic/factorial` computes the factorial of the supplied integer
  `n`.
  ([#197](https://github.com/littleredcomputer/sicmutils/pull/197)).

- Many new functions and constants exposed in `sicmutils.env` via
  [#197](https://github.com/littleredcomputer/sicmutils/pull/197):

  - `-pi` joins `pi` as a constant
  - `s:generate`, `m:generate`, `vector:generate` to generate matrices,
    structures and vectors
  - `constant-series`, from `series/constant`
  - `seq:print` and `seq:pprint`
  - `matrix-by-cols`, `row-matrix`, `v:make-basis-unit`
  - aliases for `sicmutils.function`'s `arity`, `arg-shift`, `arg-scale`
  - `dimension`, `factorial` aliased from `sicmutils.generic`
  - `derivative` aliased from `sicmutils.calculus.derivative`
  - `submatrix`, `up->column-matrix`, `down->row-matrix`,
    `row-matrix->{down,vector}`, `column-matrix->{up,vector}` aliased from
    `sicmutils.matrix`
  - `D-numeric` from `sicmutils.numerical.derivative`
  - `brent-min`, `brent-max`, `golden-section-min`, `golden-section-max`
  - `nelder-mead`
  - `sum` from `sicmutils.util.aggregate
  - `kind-predicate` from `sicmutils.value`


src/sicmutils/env.cljc
```diff
@@ -109,19 +111,30 @@
     (apply d/partial selectors)
     (apply core-partial selectors)))
 
-(def m:transpose matrix/transpose)
```
sritchie, line 112: generic transpose does this now.

src/sicmutils/expression/compile.cljc
```diff
@@ -101,9 +101,9 @@
    'tanh #(Math/tanh %)
    #?@(:cljs
        ;; JS-only entries.
-       ['acosh #(Math.acosh %)
-        'asinh #(Math.asinh %)
-        'atanh #(Math.atanh %)])})
+       ['acosh #(Math/acosh %)
```
sritchie, line 104: whoops!

src/sicmutils/matrix.cljc
```diff
@@ -41,10 +42,13 @@
 (deftype Matrix [r c v]
   v/Value
   (zero? [_] (every? #(every? v/zero? %) v))
-  (one? [_] false)
-  (identity? [_] false)
+  (one? [m] (identity? m))
```
sritchie, line 45: lots of new documentation and a few new functions in this namespace, all with generative tests.

src/sicmutils/matrix.cljc
```diff
@@ -330,42 +427,49 @@
...
+  (s/up* (map (fn [m]
+                (if (matrix? m)
+                  (->structure m s/down s/up false)
+                  m))
+              s)))
 
-(defn ^:private mul
+(defn- mul
   "Multiplies the two matrices a and b"
-  [^Matrix a ^Matrix b]
```
sritchie, line 337: I was also able to remove all type hints from these functions, so at least we won't error out if someone tries to pass some other type (and can return a sensible error).

src/sicmutils/matrix.cljc
```diff
@@ -631,31 +817,32 @@
...
-  (g/inner-product (row-matrix->up a)
-                   (row-matrix->up b)))
+  (g/inner-product (row-matrix->vector a)
+                   (row-matrix->vector b)))
 
 (defmethod g/inner-product [::column-matrix ::column-matrix] [a b]
   (g/inner-product (column-matrix->up a)
                    (column-matrix->up b)))
 
 (defmethod g/cross-product [::row-matrix ::row-matrix] [a b]
```
sritchie, line 841: @littleredcomputer , the original doesn't allow a cross product between rows. What do you think? does this matter? Should we deprecate this?

src/sicmutils/structure.cljc
```diff
@@ -353,31 +353,54 @@
 
 ;; ## Constructors
 
+(defn make
+  "Generate a structure with the supplied orientation, given some sequence `xs`"
+  [orientation xs]
+  (let [xs (if (vector? xs) xs (into [] xs))]
+    (->Structure orientation xs)))
+
+(defn up*
```
sritchie, line 362: I added these to remove the need to call `apply` if we already have a sequence.

test/sicmutils/generators.cljc
```diff
@@ -179,8 +180,9 @@
 (defn square-matrix
   ([n] (square-matrix n ratio))
   ([n entry-gen]
-   (gen/fmap #(apply m/by-rows %)
-             (gen/vector (gen/vector entry-gen n) n))))
+   (gen/fmap m/by-rows*
```
sritchie, line 183: nice removal of apply example...

test/sicmutils/matrix_test.cljc
```diff
@@ -418,7 +507,7 @@
   ;; methods, since it gets dispatched as square.
   (checking "dot-product" 100 [s (sg/up1 sg/real 2 10)]
             (let [col (m/up->column-matrix s)
-                  row (m/up->row-matrix s)]
+                  row (m/down->row-matrix (g/transpose s))]
```
sritchie, line 510: these had to change a bit to accomodate the new rigor.

src/sicmutils/matrix.cljc
```diff
@@ -41,18 +42,12 @@
 (deftype Matrix [r c v]
   v/Value
   (zero? [_] (every? #(every? v/zero? %) v))
-  (one? [_] false)
-  (identity? [_] false)
+  (one? [m] false)
+  (identity? [m] (identity? m))
   (zero-like [this] (fmap v/zero-like this))
-  ;; TODO: one-like/identity-like should use a recursive copy to find the 1 elements
```
sritchie, line 47: I'm not sure that I highlighted it in the PR description, but I took care of this too.

src/sicmutils/matrix.cljc
```diff
@@ -276,52 +294,140 @@
...
-  {:pre [(not-empty es)]}
-  (->Matrix (count es) 1 (vec (for [e es] [e]))))
+  "Maps `f` over the elements of the matrix `m` returning a new matrix of the same
+  dimensions as `m`."
+  [f m]
+  (->Matrix (num-rows m)
+            (num-cols m)
+            (mapv #(mapv f %) m)))
+
+(defn fmap-indexed
```
sritchie, line 322: this is new too.

src/sicmutils/function.cljc
```diff
@@ -66,6 +66,32 @@
...
+  ((arg-shift square 3) 4) ==> 49
+  ((arg-shift square 3 2 1) 4) ==> 49"
+  [f & shifts]
+  (let [shifts (concat shifts (repeat 0))]
+    (-> (fn [& xs]
+          (apply f (map g/+ xs shifts)))
+        (with-meta {:arity (arity f)}))))
+
+(defn arg-scale
+  "Takes a function `f` and a sequence of `shifts`, and returns a new function
```
littleredcomputer, line 83: Perhaps we should call them "scales" or "factors" in this context

src/sicmutils/matrix.cljc
littleredcomputer: I had not caught this: thanks!

src/sicmutils/matrix.cljc
littleredcomputer: sweet

src/sicmutils/matrix.cljc
littleredcomputer: I'd leave it alone for now. What we really want is to generalize this maybe if we take up more of what's going on in FDG, or we could approach this from the geometric algebra perspective. I'm not sure if there's an argument for having cross only apply to up tuples, but it doesn't bother me.

littleredcomputer (approved):
Stronger every day!

src/sicmutils/function.cljc
sritchie: done!

PR #199, opened 2020-12-11
Make sicmutils SCI consumable

kloimhardt:
@sritchie please describe the first steps

### DIFF
--- a/src/sicmutils/expression/compile.cljc
+++ b/src/sicmutils/expression/compile.cljc
@@ -22,7 +22,7 @@
   (:require #?(:cljs [goog.string :refer [format]])
             [clojure.set :as set]
             [clojure.walk :as w]
-            [sci.core :as sci]
+            [sci.core :as sci] ;; we will start here
             [sicmutils.expression :as x]
             [sicmutils.generic :as g]
             [sicmutils.structure :as struct]


sritchie:
Hey @kloimhardt , I'm so happy that you're interested here!!

The section you flagged is related, but not quite the big task.

To use the SICMUtils functions in an SCI environment, we need to declare everything we want to use in an SCI "environment" map.

We build this `sci-context` in the [namespace you've flagged](https://github.com/littleredcomputer/sicmutils/blob/master/src/sicmutils/expression/compile.cljc#L406) for the more simple case of compiling down numeric functions, and forcing them to use the native `+`, `-`, etc.

For a first pass, I think we need to get everything available in `sicmutils.env` and build a map of symbol => definition, just like we've done in [`sicmutils.expression.compile`](https://github.com/littleredcomputer/sicmutils/blob/master/src/sicmutils/expression/compile.cljc#L406).

Some tips:

- I have a function that does something similar here, in [`bootstrap-repl!`](https://github.com/littleredcomputer/sicmutils/blob/master/src/sicmutils/env.cljc#L63):

```clojure
(defmacro bootstrap-repl!
  "Bootstraps a repl or Clojure namespace by requiring all public vars from
     sicmutils.env. From (This will only work at a repl in Clojurescript.)

  TODO add support for `refer-macros` in Clojurescript
  TODO add rename, exclude support."
  []
  `(require '~['sicmutils.env
               :refer
               (into [] (keys (ns-publics 'sicmutils.env)))]))
```

(This has to be a macro because `require` is a macro on the cljs side.)

It MAY be as simple as calling `(ns-publics 'sicmutils.env)`!! I think this gives you back the map you need.

Give that a try, and then see if you can use `sci` to evaluate some expressions that use different variables here.

### Macros

I expect that macros like `with-literal-functions` might need to be handled differently. You can see that @mk had to tag a fn in [his example here](https://github.com/nextjournal/nextjournal/commit/7808c157850f026245dcaf3895cbfff15b9650e7) with `^:sci/macro` metadata. I'm happy to do that here in the sicmutils codebase if we need to, but it'd be great if you could figure that out.

@kloimhardt , we can probably end up with a `sicmutils.sci` namespace that holds everything we need - context, etc. `sicmutils.sci-test` could hold compilation tests.

kloimhardt:
As this PR is superseded by #216  it can be closed.

PR #200, opened 2020-12-12
suggestions from #193, structural prototypes keep orientations + prep for 0.14.0 release

sritchie:
This PR:

- adds missing `CHANGELOG` entries
- adds new `matrix` and `orientation` generators
- adds a test that `M^t M` is always square
- modifies `map-chain` to track the orientation sequence of a train
- `basis-unit` returns a sequence, not a vector
- modifies `structure->access-chains` to store this information as metadata on the returned seq
- `structure->prototype` now generates symbols that use proper orientation-preserving indices, an upgrade from the original!
- `seq->` had a bug, fixed now
- `m/structure->` now has a proper default implementation.

Thanks for the suggestions, @littleredcomputer.


src/sicmutils/matrix.cljc
```diff
@@ -422,25 +427,36 @@
...
+  "Returns a structure generated by converting `m` into a nested structure with
+  the supplied `outer-orientation` and `inner-orientation`.
+
+  If `t?` is true, the columns of `m` will form the inner tuples. If `t?` is
+  false, the rows of `m` will form the inner tuples.
+
+  By default, if you supply a single argument (the matrix `m`), a matrix turns
+  into a single outer `::s/down` of inner columns represented as `::up`
+  structures."
+  ([m] (->structure m ::s/down ::s/up true))
```
sritchie, line 439: I've added a default here that matches what we offer elsewhere - a matrix, by default, is a `down` of `up` columns.

src/sicmutils/matrix.cljc
```diff
@@ -422,25 +427,36 @@
...
 (defn seq->
   "Convert a sequence (typically, of function arguments) to an up-structure.
+
   GJS: Any matrix in the argument list wants to be converted to a row of
   columns"
-  [s]
+  [xs]
   (s/up* (map (fn [m]
                 (if (matrix? m)
-                  (->structure m s/down s/up false)
```
sritchie, line 441: this was a bug, maybe one that I introduced... `s/down` and `s/up` are not an orientations, so this would error.

Also, I think this is not the default behavior we want (though we don't see to be using this anywhere).

This would produce a `down` of `up` COLUMNS, which is not typical. I'm erasing these and pushing the default behavior choice back to `->structure`.

src/sicmutils/structure.cljc
```diff
@@ -58,12 +58,12 @@
...
 
 (defn basis-unit
-  "Returns a basis vector of `n` 0s, with `1` in the `i`th position.
+  "Returns a basis sequence of `n` 0s, with `1` in the `i`th position.
 
   If `n` is not supplied returns an infinite sequence."
   ([i] (map (partial kronecker i)
             (range)))
-  ([n i] (into [] (take n) (basis-unit i))))
+  ([n i] (take n (basis-unit i))))
```
sritchie, line 66: Thanks for the suggestion, @littleredcomputer !

src/sicmutils/structure.cljc
```diff
@@ -501,37 +501,56 @@
...
-               ;; subtle (I'm afraid). Here is where we put
-               ;; the access chain into the new structure.
-               ;; But if we put it in as a vector, that would
-               ;; introduce a new layer of structure since
-               ;; vectors are considered up-tuples. So we
-               ;; have to turn it into a seq, which will
-               ;; forfeit structure-nature.
-               (seq chain))
-             s))
+  (when (structure? s)
```
sritchie, line 542: I add this `when` because in multiple spots, tests pass non-structures. Before my metadata addition here, the function would return `nil` (see `map-chain` above, it would get to the `seq` call with an empty vector).

`with-meta` errors on `nil == (seq chain)`, so we get equivalent behavior by wrapping everything in `when structure?`.

src/sicmutils/structure.cljc
```diff
@@ -540,12 +559,15 @@
   - a structure `s`
 
   and returns a new structure of identical shape, with symbolic entries like
-  `'x:0:1` that show their access chain."
+  `'x↑0_1` that show their access chain with proper orientations for each step."
   [name s]
   (mapr (fn [chain]
-          (let [path (join ":" chain)]
+          (let [separators (->> (meta chain)
```
sritchie, line 565: woohoo!

test/sicmutils/matrix_test.cljc
```diff
@@ -153,6 +153,34 @@
                  (sg/square-matrix n))]
             (is (m/square? m)))
 
+  (checking "fmap-indexed" 100
```
sritchie, line 156: these check behavior introduced in the previous PR.

test/sicmutils/matrix_test.cljc
```diff
@@ -304,11 +332,23 @@
   (testing "literal-matrix"
     (let [M (m/literal-matrix 'x 3 3)]
       (is (m/matrix? M))
-      (is (= M (m/by-rows ['x↑0_0 'x↑0_1 'x↑0_2]
-                          ['x↑1_0 'x↑1_1 'x↑1_2]
-                          ['x↑2_0 'x↑2_1 'x↑2_2]))
+      (is (= M (m/by-rows ['x_0↑0 'x_1↑0 'x_2↑0]
```
sritchie, line 335: read the indices, way better, right?

test/sicmutils/matrix_test.cljc
```diff
@@ -304,11 +332,23 @@
...
           "A literal matrix is a matrix populated by symbols with index refs
-          baked in."))))
+          baked in.")))
+
+  (checking "structure prototype matches matrix literal" 100
+            [[sym M] (gen/let [sym    sg/symbol
+                               rows  (gen/choose 1 10)
+                               cols  (gen/choose 1 10)]
+                       [sym (m/literal-matrix sym rows cols)])]
+            (is (= (m/->structure M)
```
sritchie, line 346: and now, as you suggested, @littleredcomputer , we match what `s/structure->prototype` gives us back for default matrices. So good!

test/sicmutils/matrix_test.cljc
```diff
@@ -322,6 +362,13 @@
             #(m/fmap - %))))))
 
 (deftest structure
+  (checking "M^tM is always square" 100
```
sritchie, line 365: @littleredcomputer your other suggested test.

test/sicmutils/matrix_test.cljc
```diff
@@ -343,33 +390,86 @@
...
+             (m/->structure M ::s/up ::s/down false)))
+      (is (= (s/up (s/down 1 4)
+                   (s/down 2 5)
+                   (s/down 3 6))
+             (m/->structure M ::s/up ::s/down true)))
+      (is (= (s/up (s/down 1 2 3)
+                   (s/down 4 5 6))
+             (m/->structure M ::s/up ::s/down false)))))
+
+  (checking "->structure laws" 100
```
sritchie, line 419: I am addicted to generative testing these days...

test/sicmutils/matrix_test.cljc
```diff
@@ -343,33 +390,86 @@
...
+
+              (if transpose?
+                (is (= (map seq (g/transpose M))
+                       (map seq S))
+                    "the structure contains the matrix columns.")
+                (is (= (map seq M)
+                       (map seq S))
+                    "the structure contains the matrix rows."))))
+
+  (checking "seq-> is id for vectors without matrices" 100
```
sritchie, line 446: tests that lock in the new behavior in `seq->`, where we properly (imo) convert matrices.

test/sicmutils/structure_test.cljc
```diff
@@ -370,17 +370,24 @@
...
+                      (fn [_ chain _] (get-in s chain)) s)))
 
-            (is (= (s/map-chain (fn [_ chain] (seq chain)) s)
+            (is (= (s/map-chain (fn [_ chain _] (seq chain)) s)
                    (s/structure->access-chains s))
                 "map-chain and structure->access-chains are equiv"))
 
   (testing "s/structure->access-chains"
-    (is (= (s/up [0] [1] [2]) (s/structure->access-chains (s/up 1 2 3))))
+    (is (nil? (s/structure->access-chains 't))
```
sritchie, line 381: showing that I preserve the old behavior.

test/sicmutils/structure_test.cljc
```diff
@@ -409,13 +416,26 @@
...
-      (is (= (s/up 'x0
-                   (s/up 'x1:0 'x1:1)
-                   (s/down 'x2:0 'x2:1)
-                   (s/up 'x3:0 'x3:1))
+    (let [s (s/up 't
+                  (s/up   'u 'v)
+                  (s/down 'r 's)
+                  (s/up   'v1 'v2))]
+      (is (= (s/up 'x↑0
+                   (s/up   'x↑1↑0 'x↑1↑1)
```
sritchie, line 424: better!!

test/sicmutils/structure_test.cljc
```diff
@@ -409,13 +416,26 @@
...
+                   (s/down 'x↑2_0 'x↑2_1)
+                   (s/up   'x↑3↑0 'x↑3↑1))
              (s/structure->prototype 'x s)))))
 
+  (checking "prototype matches literal" 100
+            [[sym lit] (gen/let [sym sg/symbol
+                                 n   (gen/choose 1 10)
+                                 o   sg/orientation]
+                         [sym (s/literal sym n o)])]
+            (is (= lit (s/structure->prototype sym lit))
```
sritchie, line 434: same check as for matrices - `structure->prototype` preserves the indices when you apply it to a literal structure.

littleredcomputer (approved):
I really love all of this, it's perfect.

PR #202, opened 2020-12-13
mild cleanups due to new generics

sritchie:
A few spots where `abs` suffices, vs `(sqrt (square ...))` etc.


littleredcomputer (approved):
LGTM

PR #207, opened 2020-12-14
add solve-linear, solve-linear-{left,right} generic functions

sritchie:
From the CHANGELOG, this PR:

  - fixes a bug where `sicmutils.function/compose` would fail when provided with
    no arguments. Now it appropriately returns `identity`.

  - adds missing implementations of `g/floor`, `g/ceiling`, `g/integer-part` and
    `g/fractional-part` for functions, both literal and abstract.

  - adds `g/solve-linear`, `g/solve-linear-left`, `g/solve-linear-right`.
    `(g/solve-linear-right a b)` returns `x` such that `a = x*b`, while
    `g/solve-linear` (and its alias, `g/solve-linear-left`) returns `x` such
    that `a*x = b`. These functions are implemented for:

    - `sicmutils.series.{Series, PowerSeries}`
    - all numeric types
    - functions, operators
    - `sicmutils.modint.ModInt`
    - `sicmutils.differential.Differential`, so you can differentiate through
      this operation


littleredcomputer:
I want to take quaternions (the book, especially the 2ed., discusses
alternatives to Euler angles, but does not set any exercises that use them.
I think it would be interesting to repeat the rigid motion exercise with
quaternions instead of rotation matrices. That would mean plumbing the
differentiation through, and learning how to extract angular velocity ->
kinetic energy etc. That might be very enlightening.

sritchie:
@littleredcomputer it's yours! that will be awesome, and so great to show off with our pending three.js integration.

sritchie:
@littleredcomputer , made an issue here with some notes: https://github.com/littleredcomputer/sicmutils/issues/208

src/sicmutils/function.cljc
```diff
@@ -90,7 +90,8 @@
   The arity of a composition is the arity of the rightmost (that is, first to be
   applied) function term in `fns`."
   [& fns]
-  (let [a (arity (last fns))]
+  (let [a (arity (or (last fns)
```
sritchie, line 93: this fix allows `(compose)` to work, just like `(comp)` does.

src/sicmutils/function.cljc
```diff
@@ -501,11 +502,15 @@
 (defunary g/invert)
 (defbinary g/div g/divide)
 (defbinary g/expt)
+(defunary g/sqrt)
 
 (defunary g/negate)
 (defunary g/negative?)
 (defunary g/abs)
-(defunary g/sqrt)
+(defunary g/floor)
```
sritchie, line 510: whoops, these were missing!

PR #212, opened 2020-12-17
build tidying

sritchie:
- remove travis.yml
- upgrade cloverage, remove ns exclusions (even though it still can't process simplify, let's see what happens)
- remove uberjar profile


littleredcomputer (commented):
LGTM

PR #213, opened 2020-12-17
Si units: a first step

teodorlu:
**Very much work-in-progress.**

See sicmutils/placeholder#44 for discussion of implementation of unit systems, SI units and unit-aware generic arithmetic operations.

I expect that unit system design and implementation will have to mature for a while before merging.


src/sicmutils/units/units.cljc
```diff
@@ -0,0 +1,118 @@
...
+  (+ (* 5 SI/meter) (* 10 SI/meter))
+  )
+
+(comment
+  (let [meter-squared (*units &meter &meter)]
+    [(.exponents meter-squared)
+     (.scale meter-squared)]))
+
+(comment
+  (= (Units. SI [0 0 0 0 0 0 0] 1)
```
sritchie, line 103: for easy prototyping, you can either

- make it a `defrecord` for now
- take a hint from the `equals` implementation on the other deftypes

src/sicmutils/units/units.cljc
```diff
@@ -0,0 +1,118 @@
...
+     (Units. SI [0 0 0 0 0 0 0] 1))
+  ;; => false
+
+  (.hashCode (Units. SI [0 0 0 0 0 0 0] 1))
+  ;; new hashcode on each call
+
+  (Units. SI [0 0 0 0 0 0 0] 1)
+  ;; can I destructure?
+  (let [unit (Units. SI [0 0 0 0 0 0 0] 1)]
```
sritchie, line 112: you have to implement `seq` if you want to do this

src/sicmutils/units/units.cljc
teodorlu: Thanks! First time I'm using `deftype`.

teodorlu: From the `Matrix` implementation, it looks like there's different protocols to implement for Java equality and Javascript equality. `:clj`: `equals` on `Object`, and `:cljs`: `-equiv` on `IEquiv`. I could map both of those to a private method that does the right thing, like `m:=`.

sritchie:
Hey @teodorlu , closing this out now to keep things tidy. Feel free to re-open when / if you want to hack!

teodorlu:
👍

I couldn't find the time to pull this through. I don't see myself being less busy moving forward. So if others want to take a shot at physical units, just go ahead.

PR #214, opened 2020-12-17
Fix dimension for row, column matrices

sritchie:
Whoops, had these flipped. Tests and a fix for row and column matrices.


littleredcomputer (approved):
LGTM

PR #216, opened 2020-12-19
Add sicmutils.env.sci

mk:
First stab at adding a sci environment. All public functions from `sicmutils.env` should work, plus the `literal-function` macro. `with-literal-functions` might work, but I haven't tested it yet. Could not yet figure out how to get `let-coordinates` and `using-coordinates` working. 

So far just tested with Clojure. Think I should at least add some basic tests before merging this.

* [x] Create sicmutils.env.sci with public fns
* [x] Add `literal-function` and `with-literal-functions` macros
* [x] Add tests
* [x] Get ` let-coordinates` and `using-coordinates` working
* [x] Test from ClojureScript

Supersedes #199.

### DIFF
--- a/src/sicmutils/env.cljc
+++ b/src/sicmutils/env.cljc
@@ -48,6 +48,7 @@
             [sicmutils.numerical.quadrature]
             [sicmutils.mechanics.lagrange]
             [sicmutils.mechanics.hamilton]
+            [sicmutils.mechanics.rigid]
             [sicmutils.mechanics.rotation]
             [sicmutils.calculus.basis]
             [sicmutils.calculus.covariant]
@@ -55,7 +56,7 @@
             [sicmutils.calculus.form-field]
             [sicmutils.calculus.manifold]
             [sicmutils.calculus.map]
-            [sicmutils.calculus.coordinate]
+            [sicmutils.calculus.coordinate :as cc]
             [sicmutils.calculus.vector-field]))
 
 #?(:clj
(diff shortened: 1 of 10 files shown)


mk:
@sritchie I made good progress here, also thanks to the generous help of @borkdude. Adding more namespaces should be fairly trivial now.

sritchie:
@mk, I'm catching up post (inter?) holidays, and this is my next focus!

mk:
@sritchie great, let me know if anything is unclear. Is the ci coverage failure a hiccup? Regarding tests I'm unsure if it's worth it to add them now. Like you said, we'd ideally run all tests in sci but that can probably come as a follow-up PR?

sritchie:
Adding a note to myself here to check on a potential issue with "vars" that I noticed in SCI... if we don't define the environment using vars (vs fns), I think no one can access docstrings inside the Nextjournal environment. But if we DO, I noticed that sci does not seem to resolve vars into their attached fns or bindings before calling a fn.

We have a bunch of machinery that looks at the actual object to determine its type for generic dispatch, so this would be a problem.

@borkdude, does this sound familiar?

borkdude:
@sritchie 

> But if we DO, I noticed that sci does not seem to resolve vars into their attached fns or bindings before calling a fn.

I'm not sure what you mean here, but if you can demonstrate this problem using a minimal repro, I'd be happy to look into it.

src/sicmutils/env.cljc
```diff
@@ -83,10 +84,15 @@
      `(af/literal-function ~f ~sicm-signature)))
   ([f domain range] `(af/literal-function ~f ~domain ~range)))
 
-(defmacro with-literal-functions
-  [& args]
+(defmacro with-literal-functions [& args]
   `(af/with-literal-functions ~@args))
 
+(defmacro let-coordinates [& args]
```
sritchie, line 90: @littleredcomputer , I realized that these hadn't been exposed to self-hosted cljs, ie, to maria... so, adding them here vs the import below.

sritchie:
@mk , thank you for this!! I'll merge once the tests pass and get the CHANGELOG updated.

PR #218, opened 2020-12-26
Make arity a protocol, move numerical? to own protocol (autodiff #1)

sritchie:
In addition to the CHANGELOG entries that I'll mirror below, This PR:

- adds `sicmutils.util.stream/separatev`, which runs a parallel `filterv` and `filterv` on the predicate's complement
- fixes a matrix bug with clashing symbols

The changelog entry lists the rest of the changes:

- `D` (or `sicmutils.generic/partial-derivative`) applied to a matrix of
  functions now takes the elementwise partials of every function in the matrix.
  (#218)

- `sicmutils.function/arity` is now a protocol method, under the
  `sicmutils.function/IArity` protocol (#218). In addition to functions, `arity`
  now correctly responds to:

    - `sicmutils.matrix/Matrix`: calling `arity` on a matrix assumes that the
      matrix has function elements; the returned arity is the most general arity
      that all functions will respond to.
    - `sicmutils.operator/Operator`: returns the arity of the operator's wrapped
      function.
    - `sicmutils.series/Series`: `arity` on a `Series` assumes that the series
      contains functions as entries, and returns, conservatively, the arity of
      the first element of the series.
   - `sicmutils.series/PowerSeries`: `arity` returns `[:exactly 1]`, since
     `PowerSeries` are currently single variable.
   - vectors, and `sicmutils.structure/Structure`: `arity` on these collections
     assumes that the collection contains functions as entries, and returns the
     most general arity that is compatible with all of the function elements.

- New functions `sicmutils.function/{get,get-in}` added that act like the
  `clojure.core` versions; but given a function `f`, they compose `#(get % k)`,
  or similar with `f`. This deferred action matches the effect of all sicmutils
  generics on functions. (#218)

- `sicmutils.function/I` aliases `clojure.core/identity` (#218).


src/sicmutils/abstract/function.cljc
```diff
@@ -28,7 +28,7 @@
             [sicmutils.structure :as s]
             [sicmutils.util :as u]
             [sicmutils.value :as v]
-            [sicmutils.calculus.derivative :as d])
+            [sicmutils.calculus.derivative :as d :refer [derivative-symbol]])
```
sritchie, line 31: this sets me up for a later PR...

src/sicmutils/calculus/derivative.cljc
```diff
@@ -493,9 +495,6 @@
 (defmethod g/partial-derivative [::struct/structure v/seqtype] [f selectors]
   (multivariate-derivative f selectors))
 
-(defmethod g/partial-derivative [::matrix/matrix v/seqtype] [f selectors]
```
sritchie, line 496: if we pass a matrix as an argument to a fn in with other matrices, it gets converted to a structure - but scmutils seems pretty explicit that if we pass a single matrix directly, we should interpret it as a matrix of fns, and return a matrix of derivatives.

src/sicmutils/function.cljc
```diff
@@ -47,6 +73,34 @@
   (let [a (arity (last fns))]
     (with-meta (apply comp fns) {:arity a})))
 
+(defn get
+  "Returns the value mapped to key, not-found or nil if key not present.
+
+  TODO note special-cased for fns."
```
sritchie, line 79: TODO - tests for this, better docstring.

src/sicmutils/function.cljc
```diff
@@ -235,67 +289,68 @@
                       (first arities)
                       (last arities)])))))
 
-(def ^:private reflect-on-arity
-  "Returns the arity of the function f.
-  Computing arities of clojure functions is a bit complicated.
-  It involves reflection, so the results are definitely worth
-  memoizing."
+(def ^{:private true
```
sritchie, line 292: just had to move the comment to an actual docstring.

src/sicmutils/function.cljc
```diff
@@ -235,67 +289,68 @@
...
+       :doc "Returns the arity of the function f. Computing arities of clojure
+  functions is a bit complicated. It involves reflection, so the results are
+  definitely worth memoizing."}
+  reflect-on-arity
   (memoize
    #?(:cljs js-arity :clj jvm-arity)))
 
-(defn arity
```
sritchie, line 246: all these cases should now be handled by the protocol.

src/sicmutils/function.cljc
```diff
@@ -235,67 +289,68 @@
...
-                                            (fail))
-             [[:between m n] [:between m2 n2]] (let [m (max m m2)
-                                                     n (min n n2)]
-                                                 (cond (= m n) [:exactly m]
-                                                       (< m n) [:between m n]
-                                                       :else (fail)))
-             [[:between m n] [:exactly k]] (if (and (<= m k)
-                                                    (<= k n))
-                                             [:exactly k]
+  ([] [:at-least 0])
```
sritchie, line 316: I gave this new arities so we can use it as a reducing fn, without an initial value.

src/sicmutils/function.cljc
```diff
@@ -309,53 +364,52 @@
   (-> (partial comp operator)
       (with-meta {:arity [:exactly 1]})))
 
+(defn coerce-to-fn
```
sritchie, line 367: this is useful later, to have accessible by `operator.cljc`.

src/sicmutils/matrix.cljc
```diff
@@ -78,47 +80,47 @@
 
        IFn
        (invoke [_ a]
-               (Matrix. r c (mapv (fn [e] (mapv #(% a) e)) v)))
+               (Matrix. r c (mapv (fn [row] (mapv #(% a) row)) v)))
```
sritchie, line 83: ugh, I think this is the last case. There was a clash between these single-variable names and the variables we have in the deftype.

src/sicmutils/matrix.cljc
```diff
@@ -215,9 +217,9 @@
 (defn- m:= [^Matrix this that]
   (and (instance? Matrix that)
        (let [^Matrix m that]
-         (and (= (.-r this) (.-r that))
```
sritchie, line 218: not sure why this wasn't triggering a reflection warning...

src/sicmutils/operator.cljc
```diff
@@ -34,10 +34,12 @@
   (zero-like [_] (Operator. v/zero-like arity 'zero context))
   (one-like [_] (Operator. identity arity 'identity context))
   (identity-like [_] (Operator. identity arity 'identity context))
-  (numerical? [_] false)
   (freeze [_] (v/freeze name))
   (kind [_] (:subtype context))
 
+  f/IArity
+  (arity [_] arity)
```
sritchie, line 41: woohoo, arity gaining some power.

littleredcomputer (commented):
sweet.

PR #219, opened 2020-12-26
modernize Operator treatment, convert operator to deftype (autodiff #2)

sritchie:
This PR:

- Converts `Operator` to a deftype, so we can call `get` and `get-in` and have it behave appropriately
- Makes the behavior of the fns consistent with the latest SICM book
- lets `fn` and `operator` correctly with `ref` in `env.cljc`
- adds a new `component` fn to env, replacing the more limited `structure/component`

More detail lives in the CHANGELOG entries:

- #219 introduces a number of changes to `Operator`'s behavior:

  - `Operator` is now a `deftype` (not a `defrecord`); the keyword lookup for
    its `:name`, `:arity`, `:context` and `:o` fields have been replaced by,
    respectively, `o/name`, `sicmutils.function/arity`, `o/context` and
    `o/procedure` functions. This change happened to allow `Operator` to
    implement protocols like `ILookup`.

  - Native `get` and `get-in` now act on `Operator`. Given an operator function
    `f`, `get` and `get-in` compose `#(get % k)`, or similar with `f`. This
    deferred action matches the effect of all sicmutils generics on functions.

  - Combining an operator and a non-operator via `+` and `-`, the non-operator
    was previously lifted into an operator that multiplied itself by the new
    operator's argument. As of #219, this "multiplication" uses the operator
    definition of multiplication - meaning, the new operator composes the
    non-operator with its argument. Where does this matter?

    Previously adding the non-operator `sicmutils.function/I` to the identity
    operator `I` would act like this:

    ```clojure
    (((g/+ o/identity f/I) f/I) 10)
    ;; => 110 == (+ 10 (* 10 10))
    ```

    Because `f/I` multiplied itself by its argument... resulting in `(* f/I f/I)
    == g/square`.

    After the change, you see this:

    ```clojure
    (((g/+ o/identity f/I) f/I) 10)
    ;; => 20
    ```

    because `f/I` composes with its argument.

  - `sicmutils.operator/identity-operator` has been renamed to
    `sicmutils.operator/identity`

  - `o/make-operator` now takes an explicit `context` map, instead of a
    multi-arity implementation with key-value pairs.

  - `Operator` now implements `g/negate`.

  - `g/cross-product` is no longer implemented for `Operator`. operators were
    introduced by GJS to act like "differential operators", can only add, negate
    and multiply (defined as composition). We will probably relax this in the
    future, and add more functions like `g/cross-product` that compose with the
    operator's output; but for now we're cleaning house, since this function
    isn't used anywhere.

  - In this same spirit, `Operator` instances can now only be divided by scalars
    (not functions anymore), reflecting the ring structure of a differential
    operator.

- `sicmutils.env/ref` now accepts function and operators (#219). `(ref f 0 1)`,
  as an example, returns a new function `g` that acts like `f` but calls `(ref
  result 0 1)` on the result.

- The slightly more general `sicmutils.env/component` replaces
  `sicmutils.structure/component` in the `sicmutils.env` namespace (#219).
  `((component 0 1) x) == (ref x 0 1)`.


src/sicmutils/operator.cljc
```diff
@@ -120,65 +185,133 @@
...
+	            (context o)))
+
+(defn- o:-
   "Subtract one operator from another. Produces an operator which computes the
   difference of applying the supplied operators."
   [o p]
-  (->Operator #(g/- (apply o %&) (apply p %&))
-              (f/joint-arity [(:arity o) (:arity p)])
-              `(~'- ~(:name o) ~(:name p))
+  (->Operator #(g/sub (apply o %&) (apply p %&))
```
sritchie, line 257: I switched these to `g/sub` etc from `g/-` since we know, in these cases, that we're going to hit the binary case. So this skips the double dispatch.

src/sicmutils/operator.cljc
```diff
@@ -120,65 +185,133 @@
...
-              `(~'* ~(:name o) ~(:name p))
-              ;; Since operator p is applied first, it determines the type/arity
-              ;; of the composition.
-              (:context p)))
-
-(defn ^:private o*f
+  (->Operator (f/compose o p)
+              (arity p)
+              `(~'* ~(name o) ~(name p))
+              ;; TODO this seems fishy... why not unite them?
```
sritchie, line 283: I think this is a problem, @littleredcomputer , but we do get a test failure without it... I can run the tests again to see which one, noticed it when I was first writing this stuff.

src/sicmutils/operator.cljc
```diff
@@ -120,65 +185,133 @@
...
-              `(~'* ~(:name o) ~f)
-              (:context o)))
-
-(defn ^:private f*o
-  "Multiply an operator by a non-operator on the left. The
-  non-operator acts on its argument by multiplication."
-  [f o]
+                (apply o (map (fn [g] (g/mul f g)) gs)))
+              (arity o)
+              `(~'* ~(name o) ~(v/freeze f))
```
sritchie, line 303: a little catch I didn't mention in the CHANGELOG.

src/sicmutils/operator.cljc
littleredcomputer: cool.

src/sicmutils/operator.cljc
littleredcomputer: are you referring to `(context p)`? This might be more important in FDG-land, which I will have to re-immerse myself in at some point

littleredcomputer (approved):
really nice and I appreciate the documentation.

src/sicmutils/operator.cljc
```diff
@@ -96,21 +111,72 @@
...
-  (make-operator identity 'identity))
+(defn procedure
+  "Returns the backing procedure of the supplied [[Operator]]. Errors if a
+  non-[[Operator]] is supplied."
+  [op]
+  (if (operator? op)
+    (.-o ^Operator op)
+    (u/illegal (str "non-operator supplied: " op))))
+
+(defn arity
```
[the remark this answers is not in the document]
sritchie: The `arity` implementations? Yeah, piling on to the protocol from the first PR; if we decide we don't care about enforcing anything specific with arity we could back all of that out, but I think it is a good guard!

PR #220, opened 2020-12-26
Add vector-set, slight tidying in structure (autodiff #3)

sritchie:
This PR adds a "vector set" implementation that we'll use in #221 to implement `Differential`. A "vector set" is a vector of distinct, sorted elements, used to represent a set more cheaply than an actual set.

Rather than make a `deftype` and make this a real thing that matches the set interface (and perhaps discover in the process that Clojure has already optimized small sets to work just as fast, gulp!), vector sets you make with `make` stay correct if you stay inside the supplied API.

NOTE for @littleredcomputer - a `(sorted-set)` is actually very fast for conj, disj, set operations... but slower to return its greatest element.

Also, from the CHANGELOG:

- New single-arity case for `sicmutils.structure/opposite` returns an identical
  structure with flipped orientation (#220). acts as `identity` for
  non-structures.

NOTE: for reference, this idea is also called a ["flat set" in boost](https://www.boost.org/doc/libs/1_51_0/doc/html/container/non_standard_containers.html#container.non_standard_containers.flat_xxx). Technically a Clojure vector-based set is NOT flat, since vectors are implemented with a hash trie... whatever. Once we start benchmarking maybe we can replace this with a set.


src/sicmutils/util/vector_set.cljc
```diff
@@ -0,0 +1,130 @@
+;;
```
littleredcomputer, line 1: for optimization of differential tags? I wonder if it is worth specializing it to numbers

littleredcomputer (approved):

PR #221, opened 2020-12-28
New, extensible Differential implementation (autodiff #4)

sritchie:
This new implementation of `Differential` is quite similar to the old, with some performance improvements born of lots of staring.
 
I was also able to get the bug fix in that I mentioned to you, @littleredcomputer - I describe this at length in the tests for #223 , but this was a newer case of perturbation confusion that's handled in the `extract-tangent` implementation here.

The protocol that this PR adds sets us up for the "Amazing Bug" fix in #223.

Additional notes from the CHANGELOG:

- New, literate `Differential` implementation lives at at
  `sicmutils.differential` (#221) (see [this
  page](https://samritchie.io/dual-numbers-and-automatic-differentiation/) for a
  readable version.) Notable changes to the original impl at
  `sicmutils.calculus.derivative` include:

  - We've changed our terminology dlefrom GJS's `finite-part`,
    `infinitesimal-part`, `make-x+dx` to the more modern `primal-part`,
    `tangent-part`, `bundle-element` that the Automatic Differentiation community has
    now adopted.

  - A new `sicmutils.differential.IPerturbed` protocol makes it possible to
    extend the Automatic Differentiation (AD) system to be able to handle
    different Functor-shaped return values, like Java or JS lists and objects.
    See the [cljdoc page on Automatic
    Differentiation](https://cljdoc.org/d/sicmutils/sicmutils/CURRENT/doc/calculus/automatic-differentiation)
    for more detail.

  - `sicmutils.differential/{lift-1,lift-2,lift-n}` allow you to make custom
    operations differentiable, provided you can supply a derivative.

  - `Differential` implements `sicmutils.function/arity`, `IFn`, and can be
    applied to arguments if its coefficients are function values.

  - New `compare` and `equiv` implementations allow `Differential` instances to
    compare themselves with other objects using only their primal parts; this
    makes it possible to use functions like `<=`, `>`, `=` to do control flow
    during automatic differentiation. (Use `compare-full` and `eq` if you want
    to do full equality comparisons on primal and tangent components.)

  - related, `g/abs` is now implemented for `Differential` instances, making
    this function available in functions passed to `D`.

  - proper `numerical?`, `one?` and `identity?` implementations. The latter two
    only respond `true` if there are NO tangent components; This means that
    `one?` and `(= % 1)` will not agree.

  - The new implementation fixes a subtle bug with nested, higher order
    automatic differentiation - it's too subtle for the CHANGELOG, so please the
    "amazing" bug sections in `sicmutils.calculus.derivative-test` for proper
    exposition.

- `sicmutils.generic/partial-derivative` gains a `Keyword` extension, so it can
  respond properly to `:name` and `:arity` calls (#221).

### Notes

In theory, `lift-1`, `lift-2` and `lift-n` are general enough that you could use the whole package to augment ANY arithmetic, not just the sicmutils arithmetic, with forward mode AD. In reality, I realize that we've baked in reliance on `g/+` and `g/*` to the implementations of `Differential` addition and multiplication. If we let the user bind those, or pass those in to `lift-1` and friends and threaded them through, you could augment just the core functions, for example, without relying on the whole generic tower we've built here.

Some of the resources I used, studying up for this implementation:

- [R6RS-AD](https://github.com/qobi/R6RS-AD/blob/master/AD.ss) has an implementation of forward-mode AD that's slightly different than ours. This inspired me to make `lift-1` and friends recurse on differential inputs, so that the lifted function didn't have to ALREADY be lifted (ie be the same generic you're extending!)
- @qobi in DVL has a wonderful [description of forward and reverse mode AD implementation](https://github.com/axch/dysvunctional-language/blob/ebb18e2b435fb4de3207176b2da7f680f3fb8f46/dvl/doc/ad.txt).
- I never studied this, but here is a [Clojure port of R6RS-AD](https://github.com/log0ymxm/clj-auto-diff/blob/master/src/ad/core.clj). They use a nested-dual-number approach. This also has reverse mode via a mutable tape (note for #226 !)
- note the linear derivative, we can take the derivative of derivative...
- the `IPerturbed` protocol makes it possible to fix the "Amazing Bug" reported in [Manzyuk et al. 2019](https://arxiv.org/pdf/1211.4892.pdf) that GJS fixed in the 2019 release of scmutils. Here's a [talk on that paper](https://www.youtube.com/watch?v=FNlAhN0ei5M) if you want a clearer walkthrough... or look at the detailed tests in #223.

The bug in the paper arises from the question, addressed in my exposition a bit - How do you extend `derivative` so that its range can include functions? IE, can I take the derivative of a higher-order function?  This is a more complicated problem than allowing functors like `Vector` or `Dict` into the range, since the function expects an _input_ too.

What if you wanted to take derivatives of fns that returned core.async channels?


sritchie:
@littleredcomputer I've exported this PR as a literate document, if you prefer reading it that way: https://samritchie.io/dual-numbers-and-automatic-differentiation/

README.md
```diff
@@ -303,14 +303,8 @@ be done.
...
 
-Most of the code in the early sections of FDG will work, but this cannot
-be considered compelte.
-
-### Derivatives of nested functions
-
-Or, what is described as "Alexey's Amazing Bug" in the Scmutils source
-code (and further described by Oleksandr Manzyuk [here][OM]). This should
-be straightforward to fix but hasn't been necessary for the work so far.
+Most of the code in the early sections of FDG will work, but this cannot be
```
sritchie, line 306: we're getting there!

src/sicmutils/differential.cljc
```diff
@@ -0,0 +1,1408 @@
...
+;; of [[IPerturbed]] for functions and multimethods.
+;;
+;; All of this suggests that we need to make [[extract-tangent]] an open
+;; function that other folks can extend for other container-like
+;; types ([functors](https://en.wikipedia.org/wiki/Functor), specifically).
+;;
+;; The [[IPerturbed]] protocol accomplishes this, along with two other functions
+;; that we'll use later:
+
+(defprotocol IPerturbed
```
sritchie, line 350: @littleredcomputer: this is going to make it easy to extend `D` to be able to handle fns that returns java lists, js objects and arrays, quaternions or any other container someone wants to handle.

src/sicmutils/differential.cljc
```diff
@@ -0,0 +1,1408 @@
...
+  (exact? [_] false)
+  (kind [_] ::differential)
+
+  Object
+  ;; Comparing [[Differential]] objects using `equals` defaults to [[equiv]],
+  ;; which compares instances only using their non-tagged ('finite') components.
+  ;; If you want to compare two instances using their full term lists,
+  ;; See [[eq]].
+  #?(:clj (equals [a b] (equiv a b)))
+  #?(:cljs (valueOf [this] (.valueOf (finite-term this))))
```
sritchie, line 641: this was a big discovery for me, and triggered #236. This is all we need in CLJS for differentials to compare properly with non-differentials, using `<`, `>` and friends.

src/sicmutils/differential.cljc
```diff
@@ -0,0 +1,1416 @@
...
+;; whenever a non-[[Differential]] `x` would return true. To make this work,
+;; these operations look only at the [[finite-part]].
+;;
+;; HOWEVER! [[v/one?]] and [[v/zero?]] are examples of SICMUtils functions that
+;; are used to skip operations that we _want_ to happen, like multiplication.
+;;
+;; `(g/* x y)` will return `y` if `(v/one? x)` is true... but to propagate the
+;; derivative through we need this multiplication to occur. The compromise is:
+;;
+;; - [[v/one?]] and [[v/zero?]] return true only when ALL [[tangent-part]]s are
```
littleredcomputer, line 1063: This almost makes me prefer the names unity and nullity again (almost), since as you point out they're used to avoid operations like * (now I wonder if it's worth it). Nobody said computer algebra was easy

src/sicmutils/differential.cljc
```diff
@@ -0,0 +1,1416 @@
...
+    (defmethod generic-op signature [a b] (differential-op a b))))
+
+;; And now we're off to the races. The rest of the namespace
+;; provides [[defunary]] and [[defbinary]] calls for all of the generic
+;; operations for which we know how to declare partial derivatives.
+
+(defbinary g/add d:+)
+
+(defunary g/negate
+  (lift-1 g/negate (fn [_] -1)))
```
littleredcomputer, line 1303: You like `constantly` so I'm surprised you didn't use it here :)

littleredcomputer (commented):
This is monumental

src/sicmutils/differential.cljc
```diff
@@ -0,0 +1,1416 @@
...
+;;
+;; All of this suggests that we need to make [[extract-tangent]] an open
+;; function that other folks can extend for other container-like
+;; types ([functors](https://en.wikipedia.org/wiki/Functor), specifically).
+;;
+;; The [[IPerturbed]] protocol accomplishes this, along with two other functions
+;; that we'll use later:
+
+(defprotocol IPerturbed
+  (perturbed? [this]
```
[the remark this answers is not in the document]
sritchie: Yes, I wanted to ask you about this!

What you suggest is indeed the default implementation of `extract-tangent` and `replace-tag` in `IPerturbed`. The only place `perturbed?` is ever used is in [`sicmutils.abstract.function/literal-apply`](https://github.com/sicmutils/sicmutils/blob/master/src/sicmutils/abstract/function.cljc#L329), when we're detecting whether or not we're to take a derivative.

Before my changes here you performed this check with [`sicmutils.calculus.derivative/differential?`](https://github.com/sicmutils/sicmutils/blob/3155bd90c1c61d8994507a551cfa874fa859cf5f/src/sicmutils/calculus/derivative.cljc#L78)

GJS checks for [`differential?` in `literal-apply` in `litfun.scm`](https://github.com/Tipoca/scmutils/blob/master/src/kernel/litfun.scm#L440), but his `differential?`only checks for specific instances of the `Differential` type ([implementation here](https://github.com/Tipoca/scmutils/blob/master/src/kernel/types.scm#L322)).

I haven't stared at the literal function implementation yet; but it's worth documenting this more clearly! But that was the background for this method.

src/sicmutils/differential.cljc
```diff
@@ -0,0 +1,1416 @@
...
+;; Vectors don't care what they contain! We want the derivative of
+;; `square-and-cube` to also return a vector, whose entries represent the
+;; derivative of _that entry_ with respect to the function's input.
+;;
+;; But this implies that [[extract-tangent]] from the example above needs to
+;; know how to handle vectors and other collections; in the case of a vector `v`
+;; by returning `(mapv extract-tangent v)`
+;;
+;; The way that [[Differential]] instances interact with container types in
+;; SICMUtils makes it easy for these captures to occur all over. Whenever we
```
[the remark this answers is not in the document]
sritchie: Yes, I'm glad you called this out. two thoughts this triggers:

1. There's nothing wrong with what you're suggesting in principle, especially if we added a clear directional derivative operator (you'd kick the whole process off with `V + εW`). BUT you still can't stop someone from taking a differential like `5 + ε1` and putting it inside a collection! So you still need to handle both. (Concretely I'm going to remove the paragraph you're flagging, since the example above is enough to motivate the implementation! If someone makes `V + εW`, the `Differential` `IPerturbed` implementation will catch it. Otherwise we need a separate impl for vectors, etc.)

2. Anything that we allow into the coefficient slots of `Differential` needs all of its protocols implemented too, so that a function receiving a `V + εW` can do vector-ish things and not know it's received a `Differential`.

The second point is why I think `d/bundle-element` has to become generic. We do want users to be able to attempt to form `V + εW`, and we want to catch that and aggressively push `Differential` into the elements.

I DID implement `IFn` for `Differential`, because I thought it would make it easier to implement the Manzyuk et al. 2019 suggestion for how to allow the domain of `(D f)` to include functions (ie `g` is a function in `((D f) g)`), but lately I think that this is bad style too and we should rip out the whole arity, IFn etc mess and always use a wrapping function.

I can explain all that next time we talk!

src/sicmutils/differential.cljc
sritchie: What's happening to me!!

PR #222, opened 2020-12-28
Implement differential protocols for all Clojure collection types (autodiff #5)

sritchie:
From the CHANGELOG entry:

- #222 adds `v/Value` implementations for Clojure sequences and maps. Maps and
  vectors implement `f/Arity` and return `[:between 1 2]. `zero?` and
  `zero-like` work on sequence entries and map values. Maps can specify their
  `v/kind` return value with a `:type` key, and some of the calculus
  implementations do already make use of this feature. `g/partial-derivative` on
  a Clojure Map passes through to its values.

- #222 implements `d/IPerturbed` for Clojure maps, vectors and sequences;
  all are now valid return types for functions you pass to `D`.

- #222 also implements `d/IPerturbed` for SICMUtils `Matrix`, `Structure`,
  `Series`, `PowerSeries` and `Operator`.


test/sicmutils/structure_test.cljc
```diff
@@ -105,7 +105,6 @@
         `g/sin` has arity [:exactly 1]."))))
 
 (deftest arity-tests
-  (arity-check identity "vector")
```
sritchie, line 108: I had mistakenly implemented `arity` for vectors to go check the arity of its contained functions. But this is wrong, because if you try to apply a vector (vs an `up`), Clojure has already provided an `IFn` implementation that simply looks up entries!

PR #223, opened 2020-12-28
Switch derivative over to new Differential impl, solve "amazing bug" (autodiff #6)

sritchie:
This PR installs the new implementation of `Differential` into `sicmutils.calculus.derivative`. This fixes Alexey's Amazing Bug! I've also added lots of tests showing various versions of the Amazing Bug and explaining what the heck was going on.

Also:

- Adds proper `v/freeze` and `g/simplify` implementations for `Differential` (by pushing these actions into their coefficients).
- `g/simplify` implementation for Clojure maps
- tests caught a bug in my `Operator` implementation of `IPerturbed` (now fixed and tested)
- added fixes for a couple of flaky tests

From the CHANGELOG:

- #223 fixes a problem where `(operator * structure)` would return a structure
  of operators instead of an operator that closed over the multiplication.
  `::s/structure` is now properly a `::o/co-operator`, matching its status as a
  `::f/cofunction`.

- The operator returned by `sicmutils.calculus.derivative/partial` now has a
  proper name field like `(partial 0)`, instead of `:partial-derivative` (#223).

- #223 converts the implementation of `sicmutils.calculus.derivative/D` to use
  the new `Differential` type; this fixes "Alexey's Amazing Bug" and allows `D`
  to operate on higher order functions. For some function `f` that returns
  another function, `((D f) x)` will return a function that keeps `x` "alive"
  for the purposes of differentiation inside its body. See
  `sicmutils.calculus.derivative-test/amazing-bug` for an extended example.


src/sicmutils/calculus/derivative.cljc
```diff
@@ -1,400 +1,235 @@
...
+
+  For numerical differentiation,
+  see [[sicmutils.numerical.derivative/D-numeric]].
+
+  `f` must be built out of generic operations that know how to
+  handle [[d/Differential]] inputs in addition to any types that a normal `(f
+  x)` call would present. This restriction does _not_ apply to operations like
+  putting `x` into a container or destructuring; just primitive function calls."
+  [f]
+  (let [tag (d/fresh-tag)]
```
sritchie, line 189: note - this used to be inside the function, but other implementations (r6rs-AD) keep it outside. With proper `replace-tag` and `extract-tangent` implementations, which we now have, this change is good and kills one small side effect...

src/sicmutils/calculus/derivative.cljc
```diff
@@ -426,94 +260,80 @@
...
 (defn partial
-  "Partial differentiation of a function at the (zero-based) slot index
-  provided."
+  "Returns an operator that, when applied to a function `f`, produces a function
+  that computes the partial derivative of `f` at the (zero-based) slot index
+  provided via `selectors`."
   [& selectors]
   (o/make-operator #(g/partial-derivative % selectors)
-                   :partial-derivative))
+                   `(~'partial ~@selectors)))
```
sritchie, line 327: Noting this fix for the name of the returned operator!

src/sicmutils/calculus/derivative.cljc
littleredcomputer: What a great catch. It makes sense when you think about it.

src/sicmutils/structure.cljc
```diff
@@ -826,30 +827,11 @@
...
-(defmethod g/mul [::v/scalar ::structure] [a b]
-  (scalar*structure a b))
-
-(defmethod g/mul [::structure :sicmutils.operator/operator] [a b]
-  (structure*scalar a b))
-
-(defmethod g/mul [:sicmutils.operator/operator ::structure] [a b]
-  (scalar*structure a b))
-
-(defmethod g/mul [::structure :sicmutils.calculus.derivative/differential] [a b]
```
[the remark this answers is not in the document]
littleredcomputer: as it should. brilliant.

src/sicmutils/abstract/function.cljc
```diff
@@ -240,15 +241,14 @@
...
-        maxtag (->> v flatten d/max-order-tag)
-        ve (->> v (s/mapr #(d/without-tag maxtag %)) seq)
-        dv (->> v (s/mapr #(d/with-tag maxtag %)))]
-    (d/canonicalize-differential
-     (d/dx+dy (apply f ve)
-              (reduce d/dx+dy (map (fn [partialx dx]
-                                     (d/dx*dy (apply partialx ve) dx))
-                                   (flatten (make-partials f v))
-                                   (flatten dv)))))))
+        maxtag (apply d/max-order-tag (flatten v))
```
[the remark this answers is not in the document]
littleredcomputer: You know, it just occurred to me that one of the things to appreciate about this change of yours is that it brings the implementation closer to the literature that is accumulating around this topic in the FP community and that is really valuable!

src/sicmutils/calculus/derivative.cljc
```diff
@@ -426,94 +310,80 @@
...
+;; wrapped in the machinery provided by [[multivariate]]; this allows the same
+;; operator to serve functions of:
+;;
+;; - a single numerical input
+;; - a single structural input
+;; - multiple numerical OR structural inputs
+;;
+;; NOTE: The reason that this implementation is also installed
+;; for [[s/Structure]] is that structures act as functions that apply their args
+;; to every (functional) entry. Calling `(multivariate structure selectors)`
```
littleredcomputer, line 340: I'm not 100% sure what the alternative would look like here...

test/sicmutils/calculus/derivative_test.cljc
```diff
@@ -745,3 +778,547 @@
           fairly accurate estimates of Math/cos at a few points."
             (is (ish? (Math/cos 1) (via-series 1)))
             (is (ish? (Math/cos 0.6) (via-series 0.6)))))))))
+
```
littleredcomputer, line 781: This is a monumental improvement

PR #226, opened 2021-01-02
Basic reverse-mode AD implementation [in progress]

sritchie:
This is modeled after the approach described in the [dysvunctional language docs](https://github.com/axch/dysvunctional-language/blob/master/dvl/doc/ad.txt). This works now for `R^n => R` functions! It needs some work to handle nesting and all of that...

@littleredcomputer , I also took a try at an approach to share the derivatives between the `tape` and `differential` namespaces without actually making a `lift` implementation that lifts BOTH. The derivatives should certainly live somewhere, so we don't have to inline them.

Making progress!

## TODO

- [x] share derivative implementations (handled by #253)
- [ ] call the fn that calculates reverse mode a [`pullback`](https://www.juliadiff.org/ChainRulesCore.jl/stable/#The-propagators:-pushforward-and-pullback)?
- [ ] test that this can handle nesting. I'll almost certainly have to write some function to pull the tag of some differential up past a tape cell.... maybe?
- [ ] handle structural outputs too. Can I reuse my protocol definitions for `extract-tangent`, or do I have to redo everything?
- [ ] beef up the topological sort. here are some examples: [Kahn's](https://gist.github.com/alandipert/1263783) from @alandipert (described [on wiki](https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm). Note that we almost certainly want to break ties lexicographically using the gensyms (OR the active tag stack!), look into this.
- [ ] note the idea from ChainRulesCore.jl about thunking. If you want to potentially not use ALL of the derivatives, then you might use a clojure `delay` to wrap some of the partials that take a bit of work. Look at the [julia docs](http://www.juliadiff.org/ChainRulesCore.jl/v0.6/api.html#ChainRulesCore.Thunk) to see how they build this in. Would we have to extend the generic arithmetic to delays?

## NOTES:
- Based on my investigations in #262 , we need to take some care there. We have some code to check if a `differential?` is incoming; that is obviously going to fail for reverse mode.
- the tie-breaking is only going to work if derivative itself uses `with-active-tag`, so let GJS know that we do that too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+- #253 adds proper `:arglists` metadata for all generic functions, and moves the
+  derivative implementations (where relevant) onto the metadata of generic
+  functions. You can access these by calling `(<generic-function> :dfdx)` or
+  `(<generic-function> :dfdy)`, depending on whether the generic is unary or
+  binary. #253 also changes the name of macro
+  `sicmutils.generic/def-generic-function` to `sicmutils.generic/defgeneric`.
+
 - #224 adds new `Div`, `Grad`, `Curl` and `Lap` operators in
   `sicmutils.calculus.derivative` and installs them into `sicmutils.env`. #224
   also removes the `g/transpose` implementation for `Operator` instances, and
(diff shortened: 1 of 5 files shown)


adamhaber:
Another thing we might want to look into is implementing [checkpointing](https://pytorch.org/docs/stable/checkpoint.html). The motivation here is that a lot of physical computations end up as differential equations, and checkpointing can be quite useful there. I think we could use the particle example in #385 to compare the checkpointed-vs-vanilla implementations.

sritchie:
Moved to Emmy.

PR #230, opened 2021-01-07
Fix tests for sci now that 'user isn't default

sritchie:
Whoops, I removed the default `'user` namespace inclusion from the context, forcing an error here. Quick fix, plus changelog entries.

### DIFF
--- a/src/sicmutils/expression/compile.cljc
+++ b/src/sicmutils/expression/compile.cljc
@@ -19,11 +19,13 @@
 
 (ns sicmutils.expression.compile
   "This namespace compiles generic functions down into fast, native functions."
+  (:refer-clojure :exclude [gensym])
   (:require #?(:cljs [goog.string :refer [format]])
             [clojure.set :as set]
             [clojure.walk :as w]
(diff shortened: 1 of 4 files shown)


src/sicmutils/expression/compile.cljc
```diff
@@ -243,6 +245,9 @@
 ;;   out, since they never appear in this form (since they contain smaller
 ;;   subexpressions).
 
+(def gensym
```
sritchie, line 248: @mk, you caught a heisenbug for us! Add these new tests pushed this everything into JUST the right shape such that one of the tests tried to compile a function that generated intermediate bindings of different lengths.

`(gensym)` returned `G__72` for the first binding, but later bindings generated all the way up to `G__102`... and `G__102` sorts before `G__72`, so the final generated let binding blew up the tests!

The solution is to use a gensym we already had in the library, that generates symbols of the same length (sure to sort correctly).

sritchie: @littleredcomputer you might enjoy this heisenbug fix too.

PR #231, opened 2021-01-08
Expose sci namespaces and re-add bindings

mk:
A small follow-up to #216 and #230.

Makes the default sci context more batteries-included but allows consumers that are only interested in the sci namespace map to access that directly.


sritchie:
Ah, I forgot that bindings is an alias for the deeper ‘user key in the namespace map. Excellent change. Thanks @mk !

PR #232, opened 2021-01-09
Change compile-univariate-fn to compile-fn, handles n-ary fns

sritchie:
This PR modifies our function compiler to be able to handle functions of `n` arguments.

I also noticed that we still hadn't turned on the ability for Clojurescript to compile functions passed to `definite-integral`! This is turned off by default; let's see if any tests break because of this.

- [ ] make a ticket to handle structural args

Still needs tests, but would love a comment, @littleredcomputer!


littleredcomputer (approved):
Nice.

PR #235, opened 2021-01-12
Change multiplication order in lift-2 to match taylor series expansion (autodiff #7.5)

sritchie:
If we can get the simplifier to sort products and sums into a canonical (lexicographic) order, then this change would have no effect... and I'm not sure it matters anyway, given that we're using this stuff to multiply real numbers and symbolic expressions.

But I wanted to get the `lift-2` internal multiplication to match all of the other AD implementations I've found, and to match the text description of the multivariable power series.


littleredcomputer (commented):
I think things like this are worth doing and maybe it will matter for non-commutative things like quaternions?

There's nothing I hate more than when some change reshuffles elements in a sum or product post-simplification ...

sritchie:
@littleredcomputer agreed!

PR #236, opened 2021-01-13
Better compare implementations for numeric tower (autodiff #3.5)

sritchie:
The autodiff PRs exposed some weaknesses with our ability to compare numbers to each other in Clojurescript. I caught some of this stuff after I realized that by using `sg/real` I would flex a lot of the cross-type comparisons.

I also discovered `valueOf` in Javascript, which allows us to compare, say, a wrapped `Differential` or `Expression` instance (holding a number) with a native number, using the native comparison operators!!


sritchie:
@littleredcomputer , merging this and a few of the less-brainy ones to drain the queue a bit.

PR #238, opened 2021-01-16
Convert abstract fn to deftype (#7.25)

sritchie:
- #238 converts `sicmutils.abstract.function/Function` from a `defrecord` to a
  `deftype`, fixing a subtle bug where (empty f) was getting called in a nested
  derivative test.


src/sicmutils/env/sci.cljc
```diff
@@ -64,6 +64,7 @@ right to expose to a user by default. Each value is the sym->var map for the
 corresponding namespace."}
   ns-map
   {'sicmutils.env                              (ns-publics 'sicmutils.env)
+   'sicmutils.differential                     (ns-publics 'sicmutils.differential)
```
sritchie, line 67: forgot to do this in a previous PR!

test/sicmutils/calculus/derivative_test.cljc
```diff
@@ -1265,6 +1265,18 @@
                 (cont (fn f1 [y] (a x y))
                       (fn f2 [g] (g x)))))]
 
+      ;; Calling f1 or f2 separately work as expected:
+      (is (= '(((partial 0) a) t t)
+             (v/freeze
+              (((D f) 't) (fn [f1 _] (f1 't)))))
+          "`a` received `x` as its first arg, so we see `(partial 0)`")
+
+      (let [g (af/literal-function 'g)]
```
sritchie, line 1274: this was the test that was throwing an error before this change.

PR #239, opened 2021-01-18
Tidy multivariate function in derivative.cljc

sritchie:
I broke this implementation out so I can share some of the logic with a "directional derivative" function and friends. It's easier now for me to understand what was happening, and how to support more general associative structures, but I'll need a few clear use cases before the real shape of what's going on here reveals itself.

@littleredcomputer , let me know if you think this is an improvement.


src/sicmutils/calculus/derivative.cljc
```diff
@@ -262,65 +262,129 @@
...
+;;
+;; [[jacobian]] handles this main logic. [[jacobian]] can only take a structural
+;; input. [[euclidean]] and [[multivariate]] below widen handle, respectively,
+;; optionally-structural and multivariable arguments.
+
+(defn- deep-partial
+  "Returns the partial derivative of `f` with respect to the entry in `structure`
+  at the location `path`.
+
+  `entry` defaults to `(get-in structure path)`."
```
littleredcomputer, line 274: Looking at the code, this is the only meaningful choice and you carefully avoid repeating the work of get-in because you are already mapping over `chain` at the outer level. I can't think of a better way to do it offhand

littleredcomputer (approved):
This is much better than what was there before.

src/sicmutils/calculus/derivative.cljc
sritchie: that is right… I’m also realizing that a different way to accomplish this whole thing would be make `bundle-element` generic and extend it to structures too.

then instead of pushing the `assoc-in` inside of a new derivative function, we could call

```clojure
((Dv f) input (assoc-in (v/zero-like input) path 1))
```

But you’re totally right. It felt a little strange to leave the 1-extra-argument version open; it’s perfect for `map-chain` usage but it’s not useful (and we shouldn’t encourage using it) with a different entry.

PR #240, opened 2021-01-18
Fix bug with multiple arities on the JVM

### LINKED ISSUE #203: multi-arity functions break "arity" on the JVM
`sicmutils.function/arity` currently fails on calls like

```clojure
(defn cake
  ([x] x)
  ([x y] x))

(arity cake)
Execution error (IllegalArgumentException) at sun.reflect.GeneratedConstructorAccessor2/newInstance (REPL:-1).
arity? sicmutils.function$cake@63b30ba3 {:invoke [[:invoke 2] [:invoke 1]]}
```

This works in CLJS due to this section: https://github.com/littleredcomputer/sicmutils/blob/master/src/sicmutils/function.cljc#L206

We want to make the JVM version work the same by handing `(:invoke facts)` with multiple entries here: https://github.com/littleredcomputer/sicmutils/blob/master/src/sicmutils/function.cljc#L145

sritchie:
This PR fixes #203 and turns on support for multi-arity functions in `f/arity`. As a bonus, this fixes a special-cased test in derivative-test!


littleredcomputer (commented):
cool. (you're a stickler for this sort of thing: should we add a new category of arity for the case of a range-with-a-hole? It would be a little annoying to work out all the combinations for something that "never happens" with the sorts of function SICM works with (one could also argue that my `comp` hack is worse)

sritchie:
@littleredcomputer Haha, it's crossed my mind! But maybe we should wait until someone hits a place where they need this. We've been getting a bit better about having arities follow functions around.

The Clojure Way might be to do defer these to runtime errors... I'm not sure. But if you feel strongly let's get a ticket down and make it happen!

PR #241, opened 2021-01-19
Modify `render` to render non-simplified expressions, literals => string

sritchie:
This PR adjusts the infix renderers to:

- `v/freeze` the input
- ensure that the output is a string.

The former allows us to pass unsimplified (and unfrozen) `Expression` instances into `->infix` or `->TeX`. The latter makes it a bit easier to depend on the renderers from Nextjournal (cc @mk ).

CHANGELOG note:

- `->infix`, `->TeX` and `->JavaScript` in `sicmutils.expression.render` can now
  accept unfrozen and unsimplified `Expression` instances (#241). This makes it
  a bit more convenient to use `->infix` and `->TeX` at the REPL, or in a
  Notebook environment. Additionally, the return values of renderers are always
  coerced to strings. (Previously, `(->infix 10)` would return a number
  directly.)


sritchie:
@littleredcomputer , there may be a cleaner way to do this - I had to wire in the ratio detection a bit deeply.

As a reminder, the reason that ratios freeze to `(/ x y)` is that in Clojurescript there is no ratio literal. If you read `(/ x y)` back in to Clojure, it'll turn into a ratio with no fuss; if you're in `sicmutils.env` you'll also get a ratio in Clojurescript.

littleredcomputer (commented):
This looks marvelous. 

Upon reading it, I wondered whether ->TeX and ->JS should be represented as protocols. Instead we have "you're going to get frozen first, and that's a protocol" so maybe that makes more sense!

sritchie:
I agree that it would be great to be able to customize rendering for different elements... @hcarvalhoalves has been doing some stuff that would make use of this too. More to think about!

PR #252, opened 2021-02-01
Only trigger tag replacement in nested calls (suggested by GJS)

sritchie:
This PR adds an `*active-tags*` stack that allows us to save work in the case of a nested derivative. `*active-tags*` is a stack containing the tags of all nested calls; in #226 , this is going to save us trouble when deciding whether a `TapeCell` should wrap a `Differential` or vice versa if both show up in a computation.

### DIFF
--- a/src/sicmutils/calculus/derivative.cljc
+++ b/src/sicmutils/calculus/derivative.cljc
@@ -153,10 +153,14 @@
   - call the underlying `f`, producing `result`
   - return `(extract-tangent result tag)`
 
-  The returned function will also remap any instance of `tag` that appears in
-  any differential argument passed to it to a private `fresh` tag, to prevent
+  If called within the scope of a function waiting for the same `tag`, the
+  returned function will remap any instance of `tag` that appears in any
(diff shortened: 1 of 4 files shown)


src/sicmutils/calculus/derivative.cljc
```diff
@@ -236,9 +252,10 @@
   x)` call would present. This restriction does _not_ apply to operations like
   putting `x` into a container or destructuring; just primitive function calls."
   [f]
-  (let [tag (d/fresh-tag)]
-    (fn [x]
-      (-> (f (d/bundle-element x 1 tag))
+  (fn [x]
+    (let [tag    (d/fresh-tag)
+          lifted (d/bundle-element x 1 tag)]
+      (-> (d/with-active-tag tag f [lifted])
```
sritchie, line 258: this is going to help us decide WHICH tag is currently in play, when we go to implement reverse-mode and need to distinguish tape from differential... more on that in #226 .

littleredcomputer (approved):

PR #253, opened 2021-02-01
Move derivative implementations to generic metadata

sritchie:
We need the derivative implementations for both reverse mode (#226 ) and forward mode AD. This PR moves the implementations to the metadata of our generic functions. @littleredcomputer , let me know if you like this change!

I've also added slightly better `:arglists` metadata for generics, and a few docstrings.

### DIFF
--- a/src/sicmutils/differential.cljc
+++ b/src/sicmutils/differential.cljc
@@ -1183,18 +1183,25 @@
   Returns a new unary function that operates on both the original type of `f`
   and [[Differential]] instances.
 
+  If called without `df:dx`, `df:dx` defaults to `(f :dfdx)`; this will return
+  the derivative registered to a generic function defined with [[g/defgeneric]].
+
   NOTE: `df:dx` has to ALREADY be able to handle [[Differential]] instances. The
   best way to accomplish this is by building `df:dx` out of already-lifted
   functions, and declaring them by forward reference if you need to."
-  [f df:dx]
-  (fn call [x]
-    (if-not (differential? x)
-      (f x)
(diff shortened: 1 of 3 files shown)


src/sicmutils/differential.cljc
```diff
@@ -1317,119 +1330,41 @@
...
-  (lift-1 g/sinh g/cosh))
-
-(defunary g/cosh
-  (lift-1 g/cosh g/sinh))
-
-(defunary g/tanh
-  (lift-1 g/tanh
-          (fn [x]
-            (g/sub 1 (g/square (g/tanh x))))))
+(defunary g/sqrt (lift-1 g/sqrt))
```
littleredcomputer, line 1357: would it be better to move the call to lift-1 to inside defunary? Which would become private since the canonical way of making a new differentiable function would be to make it answer the :dfdx and :dfdy methods. But I do like the change overall.

littleredcomputer (approved):
I'm glad you pointed this one out to me: it is swell! It has the nice property of allowing smooth adoption of new generic functions.

src/sicmutils/differential.cljc
sritchie: I think this is a good idea, with the exception of a couple of the functions like

```clojure
(defunary g/negative?
  (fn [x] (g/negative? (finite-term x))))
```

That need access to functions like `finite-term`... but that's solved by allowing the `differential-op` to be supplied optionally.

If you don't mind I'll wait to do this until I can see what reverse mode looks like. Maybe there is some nicer way to tidy functions like the above?

This is the right move you're suggesting!

PR #254, opened 2021-02-03
Doc updates, fixes from Colin's autodiff reviews

sritchie:
@littleredcomputer , I'll accumulate changes you suggest in your autodiff reviews here, for PRs that I've already merged.

- update `Differential` docs
- remove `IFn` implementation for `Differential`

There is a better way to handle function differentials than this.


littleredcomputer (approved):
LGTM

PR #257, opened 2021-02-03
Render equality symbols correctly

hcarvalhoalves:
`sicmutils.expression.render/->infix` and `sicmutils.expression.render/->TeX` now handle equality/inequality symbols (`=`, `>=`, `>`, ...) as infix. (sicmutils/placeholder#81)

### DIFF
--- a/src/sicmutils/expression/render.cljc
+++ b/src/sicmutils/expression/render.cljc
@@ -239,8 +239,8 @@
   "Converts an S-expression to printable infix form. Numeric exponents are
   written as superscripts. Partial derivatives get subscripts."
   (make-infix-renderer
-   :precedence-map '{D 9, expt 7, :apply 5, u- 4, / 3, * 3, + 1, - 1}
-   :infix? '#{* + - / expt u-}
+   :precedence-map '{D 9, expt 7, :apply 5, u- 4, / 3, * 3, + 1, - 1, = 0, > 0, < 0, >= 0, <= 0}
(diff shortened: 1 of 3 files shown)


test/sicmutils/expression/render_test.cljc
```diff
@@ -152,6 +152,32 @@
...
 
+(deftest equations
+  (is (= "x = 4" (->infix '(= x 4))))
+  (is (= "x = 4" (->TeX '(= x 4))))
+
+  (is (= "x >= 4" (->infix '(>= x 4))))
+  (is (= "x >= 4" (->TeX '(>= x 4))))
+  
+  (is (= "x <= 4" (->infix '(<= x 4))))
+  (is (= "x <= 4" (->TeX '(<= x 4))))
```
hcarvalhoalves, line 163: Maybe we should use `\leqslant`?

src/sicmutils/expression/render.cljc
```diff
@@ -340,7 +340,9 @@
                 (str "\\begin{bmatrix}"
                      body
                      "\\end{bmatrix}")))
-      'sqrt #(str "\\sqrt " (maybe-brace (first %)))}
+      'sqrt #(str "\\sqrt " (maybe-brace (first %)))
+      '<= #(s/join " \\leq " %)
```
sritchie, line 344: excellent, I think this is a good default style vs the slanted version (which honestly looks better).

PR #258, opened 2021-02-03
Avoid sans-serif symbols

hcarvalhoalves:
`sicmutils.expression.render/*TeX-sans-serif-symbols*` binding to control if symbols longer than 1 char should have `\mathsf` applied.


sritchie:
LGTM! We should probably route more options like this in via an options dictionary, but I like this and it matches the current configuration style for the `->TeX` renderer.

PR #259, opened 2021-02-03
Use pr-str inside Literal toString, allowing proper lazy sequence printing

sritchie:
Prior to this change, we saw this:

```clojure
sicmutils.env> (pr-str (literal-number (simplify (* 'x 2))))
"clojure.lang.LazySeq@8ca01a75"
```

After, it all works as you'd expect!


littleredcomputer (approved):
LGTM

PR #260, opened 2021-02-04
Render series to TeX

hcarvalhoalves:
`sicmutils.render/->TeX` will render `Series` and `PowerSeries` as an infinite series (showing the first four terms). In the case of `PowerSeries`, it will default to `ax^n`.

### DIFF
--- a/src/sicmutils/expression/render.cljc
+++ b/src/sicmutils/expression/render.cljc
@@ -294,6 +294,7 @@
 
 (def ^:dynamic *TeX-vertical-down-tuples* false)
 (def ^:dynamic *TeX-sans-serif-symbols* true)
+(def ^:dynamic *TeX-render-infinite-series* true)
 
 (defn- displaystyle [s]
   (str "\\displaystyle{" s "}"))
@@ -348,6 +349,20 @@
      (fn r [v]
        (cond (r/ratio? v)
              (str "\\frac" (brace (r/numerator v)) (brace (r/denominator v)))
+
+             (and *TeX-render-infinite-series* (vector? v) (= (first v) 'Series))
+             (brace (-> `(~'+ ~@(subvec v 1 5)) ;; frozen value
+                        ->TeX
+                        (str " + \\ldots")))
+
+             (and *TeX-render-infinite-series* (vector? v) (= (first v) 'PowerSeries))
+             (brace (-> `(~'+
+                          ~@(map (fn [a n]
+                                   `(~'* ~a ~`(~'expt ~'x ~n)))
+                                 (subvec v 1 5) ;; frozen value
+                                 (range 4)))
+                        ->TeX
+                        (str " + \\ldots")))
 
              :else
              (let [s (str v)]
--- a/src/sicmutils/expression/render.cljc
+++ b/src/sicmutils/expression/render.cljc
@@ -278,6 +278,7 @@
    "asin" "\\arcsin",
    "acos" "\\arccos",
    "atan" "\\arctan",
+   "_" "\\_"
    "..." "\\ldots"
    })
 
--- a/src/sicmutils/expression/render.cljc
+++ b/src/sicmutils/expression/render.cljc
@@ -278,6 +278,7 @@
    "asin" "\\arcsin",
    "acos" "\\arccos",
    "atan" "\\arctan",
+   "..." "\\ldots"
    })
 
 (defn ^:private brace
@@ -294,7 +295,6 @@
 
 (def ^:dynamic *TeX-vertical-down-tuples* false)
 (def ^:dynamic *TeX-sans-serif-symbols* true)
-(def ^:dynamic *TeX-render-infinite-series* true)
 
 (defn- displaystyle [s]
   (str "\\displaystyle{" s "}"))
@@ -349,20 +349,6 @@
      (fn r [v]
        (cond (r/ratio? v)
              (str "\\frac" (brace (r/numerator v)) (brace (r/denominator v)))
-
-             (and *TeX-render-infinite-series* (vector? v) (= (first v) 'Series))
-             (brace (-> `(~'+ ~@(subvec v 1 5)) ;; frozen value
-                        ->TeX
-                        (str " + \\ldots")))
-
-             (and *TeX-render-infinite-series* (vector? v) (= (first v) 'PowerSeries))
-             (brace (-> `(~'+
-                          ~@(map (fn [a n]
-                                   `(~'* ~a ~`(~'expt ~'x ~n)))
-                                 (subvec v 1 5) ;; frozen value
-                                 (range 4)))
-                        ->TeX
-                        (str " + \\ldots")))
 
              :else
              (let [s (str v)]
--- a/src/sicmutils/series.cljc
+++ b/src/sicmutils/series.cljc
@@ -73,7 +73,7 @@
   (exact? [_] false)
   (freeze [_]
     (let [prefix (g/simplify (take 4 xs))]
-      `[~'Series ~@prefix ~'...]))
+      `(~'+ ~@prefix ~'...)))
   (kind [_] ::series)
 
   Object
@@ -227,8 +227,9 @@
   (identity-like [_] identity)
   (exact? [_] false)
   (freeze [_]
-    (let [prefix (g/simplify (take 4 xs))]
-      `[~'PowerSeries ~@prefix ~'...]))
+    (let [prefix (map-indexed (fn [n a] `(~'* ~a (~'expt ~'_ ~n)))
+                              (g/simplify (take 4 xs)))]
+      `(~'+ ~@prefix ~'...)))
   (kind [_] ::power-series)
 
   Object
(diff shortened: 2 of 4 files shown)


src/sicmutils/expression/render.cljc
```diff
@@ -349,6 +350,20 @@
...
 
+             (and *TeX-render-infinite-series* (vector? v) (= (first v) 'Series))
+             (brace (-> `(~'+ ~@(subvec v 1 5)) ;; frozen value
+                        ->TeX
+                        (str " + \\ldots")))
+
+             (and *TeX-render-infinite-series* (vector? v) (= (first v) 'PowerSeries))
+             (brace (-> `(~'+
+                          ~@(map (fn [a n]
+                                   `(~'* ~a ~`(~'expt ~'x ~n)))
```
hcarvalhoalves, line 361: I had to assume the summation is in terms of `x`.

src/sicmutils/expression/render.cljc
```diff
@@ -349,6 +350,20 @@
        (cond (r/ratio? v)
              (str "\\frac" (brace (r/numerator v)) (brace (r/denominator v)))
 
+             (and *TeX-render-infinite-series* (vector? v) (= (first v) 'Series))
+             (brace (-> `(~'+ ~@(subvec v 1 5)) ;; frozen value
```
hcarvalhoalves, line 354: Is this splicing of the frozen value too hacky? Should this representation be elsewhere (e.g. in the `Series` type)?

src/sicmutils/expression/render.cljc
sritchie: My only concern here is that I think want a way to make the `x` as an explicit variable that is only about this particular sum. Is there a good way to do this in TeX? (Concretely I'm worried about the case where there is some other `x` in play; this is actually rendering `(my-series 'x)` of course.

@littleredcomputer may have thoughts too, but I bet @hcarvalhoalves you have thought this through. What do you think?

src/sicmutils/expression/render.cljc
sritchie: I think it is fine for now. There is so much tuning we might want to introduce. I've seen, for example, instead of ellipses having `O(x^7)` or something with the order of the next unprinted term in the series.

I do it in the [`v/freeze` implementation for Series](https://github.com/sicmutils/sicmutils/blob/master/src/sicmutils/series.cljc#L74-L76); I bet you could piggyback on that, and catch the `'...` and turn that into ellipses?

```clojure
(freeze [_]
    (let [prefix (g/simplify (take 4 xs))]
      `[~'Series ~@prefix ~'...]))
```

sritchie:
Hey, I think this is a great thing to be able to have. Let's do it (especially since it's behind a dynamic variable, thank you!)

You've got me thinking about how one might make a `PowerSeries` or `Series` instance that tries to keep its concise summation representation as long as possible. But that is a bigger project.

Another bigger idea here is making the whole rendering toolkit generic, so these changes would be simple. Multimethod dispatch would make it easy to build more specialized renderers that extend `TeX` by writing (derive ::my-renderer ::render/tex)`, for example, and just overriding a few of the renderers.

src/sicmutils/expression/render.cljc
hcarvalhoalves: @sritchie I had no better ideas... I believe something more sophisticated would require the `PowerSeries`  to carry some metadata about what variable is bound on the summand?

hcarvalhoalves:
> Hey, I think this is a great thing to be able to have. Let's do it (especially since it's behind a dynamic variable, thank you!)
> 
> You've got me thinking about how one might make a `PowerSeries` or `Series` instance that tries to keep its concise summation representation as long as possible. But that is a bigger project.
> 
> Another bigger idea here is making the whole rendering toolkit generic, so these changes would be simple. Multimethod dispatch would make it easy to build more specialized renderers that extend `TeX` by writing (derive ::my-renderer ::render/tex)`, for example, and just overriding a few of the renderers.

@sritchie I can try to refactor this to use dispatch later as I keep adding more use cases 👍

src/sicmutils/expression/render.cljc
sritchie: Adding this comment from Slack:

the big distinction between `Series` and `PowerSeries` is that a `PowerSeries` is actually an unapplied function of the indeterminate. When you apply it to an argument it turns into a `Series`. so `(pseries 'x)` , or any other `Series`, makes sense to render and your solution is great.

@colins suggested that we might render a `PowerSeries` with `_` as the indeterminate, to make clear that it’s an unapplied function. that feels better to me than picking a more common symbol, and also better than refusing to render a `PowerSeries` without the user applying it to turn it into a `Series`.

wdyt, @hcarvalhoalves ?

src/sicmutils/expression/render.cljc
hcarvalhoalves: Would it make sense to change the `freeze` representation of `Series` to already represent the infinite sum?

```
`(~'+ ~@prefix ~'...)
```

Then I don't need to special case `Series` in the render function (just the ellipsis).

hcarvalhoalves: See https://github.com/sicmutils/sicmutils/pull/260/commits/03d0c242770bddf001a415fda3c7892c640c77e2

src/sicmutils/expression/render.cljc
hcarvalhoalves: See https://github.com/sicmutils/sicmutils/pull/260/commits/03d0c242770bddf001a415fda3c7892c640c77e2

test/sicmutils/expression/render_test.cljc
```diff
@@ -189,6 +189,24 @@
...
+(deftest series-tests
+  (is (= "1 + 1 + 1/2 + 1/6 + ..."
+         (->infix (series/exp-series 1))))
+  (is (= "1 + 1 + \\frac{1}{2} + \\frac{1}{6} + \\ldots"
+         (->TeX (series/exp-series 1))))
+  (is (= "(k + 1) + 1 + 1/2 + 1/6 + ..."
+         (->infix (+ 'k (series/exp-series 1)))))
+  (is (= "\\left(k + 1\\right) + 1 + \\frac{1}{2} + \\frac{1}{6} + \\ldots"
+         (->TeX (+ 'k (series/exp-series 1)))))
+  (is (= "1 _⁰ + 2 _¹ + 3 _² + 4 _³ + ..."
```
sritchie, line 201: Another option for `freeze` might be:

```clojure
(freeze [_]
    (let [prefix (->> (take 4 xs)
                      (map-indexed
                       (fn [i a]
                         (g/* a (g/expt '_ i))))
                      (g/simplify))]
      `(~'+ ~@prefix ~'...)))
```

That converts, for example, `1*x` => `x`, and cancels out entries with a `0` coefficient, so the representation is slightly simpler.

The problem is that this can bring the `_^2` etc elements in front of the coefficient:

```clojure
sicmutils.series> (power-series* ['a 0 (g/square 'b) 'c])
#object[sicmutils.series.PowerSeries "(+ a 0 (* (expt _ 2) (expt b 2)) (* (expt _ 3) c) ...)"]
```

So let's stay with what you've got.

src/sicmutils/series.cljc
```diff
@@ -227,8 +227,9 @@
   (identity-like [_] identity)
   (exact? [_] false)
   (freeze [_]
-    (let [prefix (g/simplify (take 4 xs))]
-      `[~'PowerSeries ~@prefix ~'...]))
+    (let [prefix (map-indexed (fn [n a] `(~'* ~a (~'expt ~'_ ~n)))
```
sritchie, line 230: My only suggestion is that we change freeze slightly, to

```clojure
(freeze [_]
    (let [prefix (map-indexed
                  (fn [n a] `(~'* ~a ~(g/expt '_ n)))
                  (g/simplify (take 4 xs)))]
      `(~'+ ~@prefix ~'...)))
```

You can see the difference here:

```clojure
sicmutils.series> (power-series* ['a 0 (g/square 'b) 'c])
#object[sicmutils.series.PowerSeries "(+ (* a (expt _ 0)) (* 0 (expt _ 1)) (* (expt b 2) (expt _ 2)) (* c (expt _ 3)) ...)"]
sicmutils.series> (power-series* ['a 0 (g/square 'b) 'c])
#object[sicmutils.series.PowerSeries "(+ (* a 1) (* 0 _) (* (expt b 2) (expt _ 2)) (* c (expt _ 3)) ...)"]
```

The first two entries become `1` and `_`, instead of `_^0` and `_^1`.

sritchie (approved):
This looks wonderful, just one small comment / request!

src/sicmutils/series.cljc
sritchie: Actually, let's leave this out for now and keep it as you have it. Otherwise we get confusing entries like `(* 1 1)` in the constant term, since we simplify before, not after (and after causes problems for other reasons).

Merged!

sritchie:
Thanks for this, @hcarvalhoalves !

PR #262, opened 2021-02-08
Expand literal function descriptor language [in progress]

### LINKED ISSUE #54: Flesh out literal fn syntax for structured args, values to match scmutils
We do support such literal functions, but with an undocumented syntax for domain and range. This should be harmonized with the book.

sritchie:
Full port of `litfun.scm` and the function descriptor language from scmutils.

== Literal Function Descriptor Language

We support _some_ of the possible literal function types; these are type signatures like `(-> (UP Real Real) Real)`. But [`litfun.scm`](https://github.com/Tipoca/scmutils/blob/master/src/kernel/litfun.scm) has support for a richer language.

* study `litfun.scm` and port the full scope of function descriptor language, including lots of helper functions like `Lagrangian`. There are also many examples here that would make great tests. This can live in `sicmutils.abstract.function.cljc`
* there is also a "typed function" we don't have that would be interesting to study.
* Forms like `(UP* Real)` should work, with no explicit arity.
* add 0-arity of Hamiltonian back in: `([] '(-> (UP Real (UP* Real) (DOWN* Real)) Real))`
* add 0-arity of Lagrangian back in: `([] '(-> (UP Real (UP* Real) (UP* Real)) Real))`. the current impl can't handle stars with no numbers.


sritchie:
Moved to Emmy.

Issue #275, opened 2021-02-13
add `:compile?` option to `Lagrangian-action`, fix bug with cljs compile returning number


sritchie:
Three issues here:

- [x] thread a compile option through in `Lagrangian-action`: https://github.com/sicmutils/sicmutils/blob/master/src/sicmutils/mechanics/lagrange.cljc#L134-L136  (FIXED by #276)
- [x] If `sicmutils.expression.compile/compile-fn*` simplifies down to a NUMBER in Clojurescript, we want to pass that number through `js/Number`... otherwise it can survive as a `js/BigInt` and cause trouble. (FIXED by the `v/freeze` call added at some point after this ticket was created.)
- [x] `parametric-path-action` needs a compile option. But if we give it that we need to compile the Lagrangian-action function ONE time, not every time it's called! (TODO not anymore, not now that we use an actual polynomial there.)

Another error we might have is that the compiled function might have non-function symbols left in it. We want to error if this is the case. Here's a repro from Kees:

```clojure
(defn test-path [t]
  (up (+ (* 4 t) 7)
      (+ (* 3 t) 5)
      (+ (* 2 t) 1)))

(Lagrangian-action (L-free-particle 'm) test-path 0 2 {:compile? true}) 
;; =>  crashes hard
```

This currently fails with a "symbol not recognized: 'm"... which I think is appropriate for now, during compilation.

### Overview

(NOTE: These notes are good, but more general than this ticket!)

We eventually want compilation to be more like specialization to a particular environment. It happens to be numbers here. So we should collapse an expression that has:

- JUST symbols, fine since we assume symbols are provided as arguments
- functions of symbols
- bare literals of the type that we expect, like a number...
- we do NOT want to allow functions that take bare literals, since we can evaluate those.

But we also want to do common subexpression elimination BEFORE we evaluate the bare literals, so we only do a subexpression once. But then THAT might trigger a round of replacement.

Can we do this as we're extracting common subexpressions? Some extra piece of the recursion?

PR #276, opened 2021-02-13
Add integration-opts to Lagrangian-action

sritchie:
Working on this to address #275 . See #275 for a big discussion of the issue.

NOTES: no longer compiling by default for parametric-path-action... we don't need this since we have a polynomial now and it's fast.

From the CHANGELOG:

- #276 adds an `integration-opts` to
  `sicmutils.mechanics.lagrange/Lagrangian-action`. All options are passed on to
  `definite-integral`. By default, `parametric-path-action` passes `:compile?
  false`, since we do NOT want to compile the polynomial.


sritchie:
Thanks @kloimhardt for the report, I'll get this fixed up!

Issue #272, opened 2021-02-13
Modulo function does not work with symbol argument


pangloss:
Even though `modulo` is present within the library, it's not exported by the env namespace and it also seems to be only a partial implementation since this basic use case doesn't work:

```clojure
(->infix (* 2 'x)) ; => "2 x"
(->infix (sicmutils.generic/modulo 2 'x)) ;; error
```

This seems like an oversight? I noticed the same issue with `remainder`.

I looked to see if I could simply implement the missing functionality and submit a PR, but after 10 minutes of searching I still couldn't figure out where that would be implemented.

sritchie:
@pangloss it's definitely a bug that it's not exported into env - I'll get that fixed right away, thanks for the catch!

I currently only have `modulo` installed for functions, integers and `sicmutils.modint/ModInt` instances. Your question about modulo between symbols is, I think, related to sicmutils/placeholder#81 in this way (your design thoughts would be great here):

In the original `scmutils`, all "literal expressions", or symbols, are treated as (symbolic) complex numbers. There was SOME machinery in the library to specify that certain symbols were reals... but the general idea that you really want is the ability to make abstract vectors, matrices, reals, rationals, whatever, and have them only respond to the proper operations.

SO! That's why certain stuff is missing for symbols. `modulo` is (I'm pretty sure) only defined for integers, so it's not in there for real / complex symbols.

On the other hand, sicmutils/placeholder#81 is a weird case where of course we want to be able to write `(= 'a 'b)` and have that return a symbolic equality expression. But now the type is not Complex anymore! It's an abstract boolean. So what do you do? Make a new kind of "abstract boolean" probably, and then add symbolic control flow operators like `if` etc.

Anyway, just rambling now. Given that background, what do you think? Should `(modulo 2 'x)` return that form?

Quizzing @littleredcomputer too...

Thanks for reporting this!

pangloss:
I hadn't ever encountered a complex modulo but Wolfram Alpha produces results for them and I found [this reference](https://mathworld.wolfram.com/ComplexModulus.html). On the other hand, modulo is definitely useful on real numbers.

pangloss:
Without the benefit of any insight into the codebase or anything more than fragementary mathematical knowledge, would it make sense to have a "constrained type" or wrapper that could indicate that a symbol is some subset of the possible types? That could allow you to express that `(= 'a 'b)` is boolean, and that for `(mod 2 'x)`, `x` is one of `#{int real}`.

[...]

PR #278, opened 2021-02-14
Implemented modulo, remainder, floor, ceiling, integer-part, fractional-part

pangloss:
Related to #272

### Next Tasks

@pangloss, don't feel any compulsion to do all of these, but these are the tasks that this push generates.

- [ ] TeX Rendering! Notes on [fractional-part](https://math.stackexchange.com/questions/1882055/notation-for-specifying-the-fractional-part-of-a-number), and [integer-part](https://math.stackexchange.com/questions/1882055/notation-for-specifying-the-fractional-part-of-a-number), [floor](https://tex.stackexchange.com/questions/118173/how-to-write-ceil-and-floor-in-latex), [ceiling](https://tex.stackexchange.com/questions/118173/how-to-write-ceil-and-floor-in-latex).
- [ ] CHANGELOG entry
- [ ] We'll need to look up the unicode characters for the infix renderer too.
- [ ] Generative tests, lock in the "laws" for how these types interact
- [ ] Docstrings for the new methods explaining how they relate, and the choices we've made
- [x] Change the Fraction implementation to match Clojure's Numbers.java impl
- [ ] add a `v/integral?` precondition to modint's `make`
- [ ] implement these functions for `Differential`; practically this means add `:dfdx` and `:dfdy` entries to each new function
- [ ] research what an implementation of `modulo` looks like on structures and vectors. I think I found a reference to how to do this. We can make a ticket for this.


pangloss:
I'm not at all certain that integer-part and fractional-part need to be added to this system. My suspicion is they may be useful, they're present in other systems, and seem to be symmetrical to functions like `real-part`.

src/sicmutils/generic.cljc
```diff
@@ -165,15 +168,42 @@ See [[*]] for a variadic version of [[mul]]."
...
 
-(defgeneric remainder 2)
+;; for a complex, return the integer/fractional part of both parts
+(defgeneric integer-part 1)
+(defgeneric fractional-part 1)
+(defmethod fractional-part :default [a]
+  (sub a (integer-part a)))
+
+;; complex floor/ceiling does floor/ceiling on both parts
+(defgeneric floor 1)
```
sritchie, line 178: We should be able to implement all of these for `sicmutils.differential` too, with this as a template: https://math.stackexchange.com/questions/305949/derivative-of-floor-function

src/sicmutils/complex.cljc
```diff
@@ -155,6 +155,22 @@
...
 
+(defmethod g/integer-part [::complex] [^Complex a]
+  (complex (g/integer-part (.getReal a))
+           (g/integer-part (.getImaginary a))))
+
+(defmethod g/fractional-part [::complex] [^Complex a]
+  (complex (g/fractional-part (.getReal a))
+           (g/fractional-part (.getImaginary a))))
+
+(defmethod g/floor [::complex] [^Complex a]
```
sritchie, line 166: Nice! In the cljs version below we can use the native `floor` and `ceil` methods: https://github.com/infusion/Complex.js/#complex-floorplaces0

src/sicmutils/complex.cljc
```diff
@@ -155,6 +155,22 @@
 (defmethod g/tanh [::complex] [^Complex a] (.tanh a))
 (defmethod g/simplify [::complex] [a] (v/freeze a))
 
+(defmethod g/integer-part [::complex] [^Complex a]
```
sritchie, line 158: let's document what you find about how each of these pulls toward 0 or away from 0 and how they're related. Adding a low-level method like this generates a bunch of activity, so feel free to make a TODO list in the PR. If we can gather references I can help get the docs written well on the cljdoc site describing what's going on.

sritchie: OEIS has some good stuff: https://oeis.org/wiki/Integer_part

I appreciate their note about multiple possible interpretations of fractional-part for negative numbers: https://oeis.org/wiki/Fractional_part

src/sicmutils/complex.cljc
sritchie: `round` might help us too.

src/sicmutils/env.cljc
```diff
@@ -201,6 +201,8 @@
   abs
   sqrt
   quotient remainder modulo
+  floor ceiling
```
sritchie, line 204: expanding the menu, I love it!

src/sicmutils/generic.cljc
```diff
@@ -147,6 +147,9 @@ See [[*]] for a variadic version of [[mul]]."
   {:name '/
    :dfdx (fn [x] (div -1 (mul x x)))})
 
+(defmethod invert :default [a]
```
sritchie, line 150: Interesting, I already have `div` defined in terms of `invert`:

```clojure
(defmethod div :default [a b]
  (mul a (invert b)))
```

the idea being that `invert` is the actual basic thing that needs to exist to upgrade the multiplicative monoid to a group.

BUT! I also like defining one in terms of the other, so you only have to override one. (You'll get a stack overflow instead of a Not Implemented error if you don't override either, which is the problem.)

Can you see a way to detect and error easily if a type hits BOTH defaults?

Storehaus did this for get and multiGet: https://github.com/twitter/storehaus/blob/develop/storehaus-core/src/main/scala/com/twitter/storehaus/ReadableStore.scala#L170-L181

src/sicmutils/generic.cljc
```diff
@@ -165,15 +168,42 @@ See [[*]] for a variadic version of [[mul]]."
 
 (defgeneric quotient 2)
 
-(defgeneric remainder 2)
+;; for a complex, return the integer/fractional part of both parts
+(defgeneric integer-part 1)
+(defgeneric fractional-part 1)
+(defmethod fractional-part :default [a]
+  (sub a (integer-part a)))
```
sritchie, line 175: Does this do the right thing for negative numbers? I can't see that it does or doesn't, just haven't dug in to what the proper semantics are yet. (It seems like it must, since the law we want is that

```clojure
(is (= a (+ (fractional-part a) (integer-part a))))
```

 for all `a`.

src/sicmutils/numbers.cljc
```diff
@@ -58,6 +58,10 @@
 (defmethod g/magnitude [::v/real] [a] (u/compute-abs a))
 (defmethod g/div [::v/real ::v/real] [a b] (core-div a b))
 (defmethod g/invert [::v/real] [a] (core-div a))
+(defmethod g/integer-part [::v/real] [a] (int a))
+(defmethod g/floor [::v/real] [a] (int (Math/floor a)))
+(defmethod g/ceiling [::v/real] [a] (int (Math/ceil a)))
+(defmethod g/modulo [::v/real ::v/real] [a b] (mod a b))
```
sritchie, line 64: This already exists on like 165; I'd recommend changing the types below to `::v/real` and deleting this line.

src/sicmutils/numbers.cljc
```diff
@@ -58,6 +58,10 @@
 (defmethod g/magnitude [::v/real] [a] (u/compute-abs a))
 (defmethod g/div [::v/real ::v/real] [a b] (core-div a b))
 (defmethod g/invert [::v/real] [a] (core-div a))
+(defmethod g/integer-part [::v/real] [a] (int a))
```
sritchie, line 61: Let's use `long` vs `int` in these three functions. This should be fine in `cljs` too because `long` acts as identity, so there's no need to special-case.

src/sicmutils/numsymb.cljc
```diff
@@ -117,6 +117,39 @@
         (nil? args) (g/invert arg)
         :else (div arg (reduce mul args))))
 
+(defn- mod-rem [a b f sym]
```
sritchie, line 120: This is similar in spirit to [with-exactness-preserved](https://github.com/sicmutils/sicmutils/blob/master/src/sicmutils/numsymb.cljc#L46). Should we house them next to each other?

src/sicmutils/numsymb.cljc
```diff
@@ -117,6 +117,39 @@
...
+        :else (list sym a b)))
+
+(defn- modulo [a b]
+  (mod-rem a b modulo 'modulo))
+
+(defn- remainder [a b]
+  (mod-rem a b remainder 'remainder))
+
+(defn- floor [a]
+  ;; TODO: should almost-integer? be used in floor and ceiling? (dw Feb 13 2021)
```
sritchie, line 133: I don't quite see how it would help - if it's a number, you'll floor it whether it's close or not, right?

src/sicmutils/numsymb.cljc
```diff
@@ -117,6 +117,39 @@
...
+  (if (v/number? a)
+    (g/floor a)
+    (list 'floor a)))
+
+(defn- ceiling [a]
```
sritchie, line 138: Hmm, do you think we should do the `with-exactness-preserved` trick, and actually emit `(ceiling 1.2)` for example if the number is not exact? I think no here... but MAYBE for `integer-part`. The question has to do with if there are simplifications that might be able to strip these wrappers off.

I think there are not. There certainly are none now!

src/sicmutils/numsymb.cljc
```diff
@@ -407,6 +440,12 @@
    '- sub-n
    '* #(reduce mul 1 %&)
    '/ div-n
+   'modulo #'modulo
```
sritchie, line 443: Are you putting the var in the value so that the docstring etc can come along for the ride?

src/sicmutils/ratio.cljc
```diff
@@ -197,6 +197,10 @@
                         (str n) "/" (str d)
                         "\"")))))))
 
+(defmethod g/floor [Ratio] [a] (int (Math/floor a)))
```
sritchie, line 200: You shouldn't need these on the Clojure side, because the `::v/real` method will catch them. In Clojurescript, you have `floor` and `ceil` available in Fraction.js: https://github.com/infusion/Fraction.js/#fraction-ceilplaces0-16

`round` too if that helps.

sritchie (commented):
Excellent stuff! Left a bunch of thoughts that will, hopefully, help with the next push.

src/sicmutils/generic.cljc
pangloss: It'd be possible to detect that both defaults are hit with code like this:

```clojure
(def ^:dynamic *in-default-invert* false)

(defmethod invert :default [a]
  (binding [*in-default-invert* true]
    (div 1 a)))

(defmethod div :default [a b]
  (if *in-default-invert*
    (throw (ex-info "No implementation of invert or div." {:method 'div :args [a b]}))
    (mul a (invert b))))
```

But I think a better way is to query the div multimethod directly from within invert. In my instrumentation branch, I added a helper function that would facilitate that:

```clojure
(defn method-dispatched [multimethod & args]
  (get (clojure.set/map-invert (.getMethodTable multimethod))
       (.getMethod multimethod
                   (apply (.dispatchFn multimethod) args))))
```

With that you could do this, which despite being a bit unconventional I think is a very clean solution:

```clojure
(defmethod invert :default [a]
  (when (= :default (method-dispatched div 1 a))
    (throw (ex-info "No implementation of invert or div." {:method 'div :args [a b]})))
  (div 1 a))
```

Both solutions can be easily done in cljs as well.

src/sicmutils/numsymb.cljc
pangloss: The question is whether you want  `(floor 1.999...999) ; => 1`.

src/sicmutils/numsymb.cljc
pangloss: I just used the var there to make it easier to debug in the repl.

src/sicmutils/numsymb.cljc
pangloss: It seems counterintuitive that you'd ever be able to strip off floor or ceiling since they seem nonlinear.

src/sicmutils/generic.cljc
pangloss: That was my thinking. In WA, the fractional part of a negative number is negative: https://reference.wolfram.com/language/ref/FractionalPart.html

src/sicmutils/generic.cljc
pangloss: On further thought I decided the dynamic var is probably a better idea since it doesn't require the overhead of inverting the `div` method table on every call.

pangloss:
@sritchie I switched the fractional-part implementation to the one that OEIS claims is more commonly used, but given how ambiguous the definitions of some of these functions are I wonder if it'd be a good idea to make them configurable? 

Something along these lines:


```clojure
(define *fractional-part-style* floor)

(defmethod fractional-part :default [a]
  (sub a (*fractional-part-style* a)))
```

What do you think?

src/sicmutils/generic.cljc
pangloss: @sritchie I've made this change but let me know if you think it's the right approach.

src/sicmutils/complex.cljc
pangloss: @sritchie I switched the fractional-part implementation to the one that OEIS claims is more commonly used, but given how ambiguous the definitions of some of these functions are I wonder if it'd be a good idea to make them configurable? 

Something along these lines:


```clojure
(define *fractional-part-style* floor)

(defmethod fractional-part :default [a]
  (sub a (*fractional-part-style* a)))
```

What do you think?

src/sicmutils/generic.cljc
pangloss: I'll need to leave this task to you since I don't yet understand this area at all.

sritchie (commented):
This is really great. Just a couple more comments / requests and then let's get this in! I can write the CHANGELOG entry etc in a follow-up PR where I'll implement these for Differential and friends.

We should also get support for these into the TeX renderer.

I'll update the PR notes with "TODO" items for myself on what's next.

Nice work!

test/sicmutils/abstract/number_test.cljc
```diff
@@ -497,7 +509,7 @@
            (g// (g/* 10 20 'x) (g/* 'x 3 2 1)))))
 
   (testing "negate"
-    (is (= (g/+ 'x (g/- 'x) (g/+ 'x (g/negate 'x)))))
+    (is (= (g/+ 'x (g/- 'x)) (g/+ 'x (g/negate 'x))))
```
sritchie, line 512: Good catch!

src/sicmutils/complex.cljc
```diff
@@ -155,6 +155,14 @@
...
+  (complex (g/integer-part (g/real-part a))
+           (g/integer-part (g/imag-part a))))
+
+(defmethod g/fractional-part [::complex] [^Complex a]
```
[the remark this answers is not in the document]
pangloss: I've got this change pending.

src/sicmutils/numbers.cljc
```diff
@@ -192,6 +200,7 @@
      ;; native BigInt type in JS.
      (defmethod g/add [js/BigInt js/BigInt] [a b] (core-plus a b))
      (defmethod g/mul [js/BigInt js/BigInt] [a b] (core-times a b))
+     (defmethod g/modulo [js/BigInt js/BigInt] [a b] (g/modulo-default a b))
```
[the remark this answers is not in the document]
pangloss: Good catch.

src/sicmutils/numbers.cljc
```diff
@@ -227,21 +236,22 @@
          (op (js/Number a) b))
 
        (defmethod op [::v/floating-point js/BigInt] [a b]
-         (op a (js/Number b)))
+         (op a (js/Number b))))
+
+     ;; BigInt can't handle these operations natively, so we override with a
```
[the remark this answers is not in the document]
pangloss: Nope, this was mistakenly included in the above doseq.

src/sicmutils/ratio.cljc
```diff
@@ -226,6 +226,10 @@
      (defmethod g/mul [Fraction Fraction] [^Fraction a ^Fraction b] (promote (.mul a b)))
      (defmethod g/div [Fraction Fraction] [^Fraction a ^Fraction b] (promote (.div a b)))
      (defmethod g/exact-divide [Fraction Fraction] [^Fraction a ^Fraction b] (promote (.div a b)))
+
+     (defmethod g/modulo [Fraction Fraction] [a b] (rationalize (g/modulo-default a b)))
```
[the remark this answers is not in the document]
pangloss: I didn't implement modulo or remainder in terms of one another.

You're right that there is a method provided by the Fraction library to do modulo and remainder, so I've updated to use that instead of the default.

src/sicmutils/complex.cljc
pangloss: That makes sense. I've put a note in the docstring for fractional-part.

test/sicmutils/complex_test.cljc
```diff
@@ -123,6 +123,47 @@
...
+
+      (is (= (c/complex -2 2)
+             (g/modulo (c/complex 6 4) (c/complex 3 5)))))
+
+    (testing "integer-part"
+      (is (= (c/complex 1 2) (g/integer-part (c/complex 1 2))))
+      (is (= (c/complex 1 2) (g/integer-part (c/complex 1.5 2.9))))
+      (is (= (c/complex -1 -2) (g/integer-part (c/complex -1.5 -2.9)))))
+
+    (testing "fractional-part"
```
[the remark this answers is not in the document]
pangloss: Aren't those implementations too trivial to bother? I don't think it would actually provide much additional assurance. (I'm normally a fan of generative tests)

sritchie:
@pangloss thank you for all of your work here! I'll go ahead and merge this tomorrow - I want to give it a good stare once I've gotten some sleep, and copy that task list I made above over into tickets.

I should be able to get this deployed shortly upon merge.

sritchie:
Hey, just noting that my Dad-of-young-daughter schedule is fragmented enough that I may have to push this back one more day, but this is awesome and we'll get these in soon.

sritchie:
Thanks again for this, @pangloss ! I'll take a pass over docs and CHANGELOG.

src/sicmutils/numbers.cljc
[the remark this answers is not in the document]
sritchie: Nice, thanks!

test/sicmutils/complex_test.cljc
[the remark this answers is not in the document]
sritchie: You'd hope so! But then go look at some of the funky string-representation-based implementations people have for these functions, and I think you'll agree that the "trivial" test could catch those out pretty quick.

Functions with very straightforward laws are nice and understandable; it's only fair to the function to write down the relationship in executable form.

PR #279, opened 2021-02-14
Alias docstrings in aliased defs in sicmutils.env

sritchie:
From the CHANGELOG:

- Function aliases in `sicmutils.env` now properly mirror over docstrings
  and other `Var` metadata, thanks to
  [Potemkin](https://github.com/clj-commons/potemkin)'s `import-def`. This
  doesn't quite work in Clojurescript since we can't use `resolve` inside of a
  macro (special form!).

### DIFF
--- a/src/sicmutils/env.cljc
+++ b/src/sicmutils/env.cljc
@@ -40,8 +40,8 @@
   (:refer-clojure :rename {ref core-ref
                            partial core-partial
                            compare core-compare}
-                  :exclude [+ - * / zero? compare #?(:cljs partial)])
-  (:require #?(:clj [potemkin :refer [import-vars]])
+                  :exclude [+ - * / zero? compare divide #?(:cljs partial)])
+  (:require #?(:clj [potemkin :refer [import-def import-vars]])
             #?(:clj [nrepl.middleware.print])
             [sicmutils.abstract.function :as af #?@(:cljs [:include-macros true])]
             [sicmutils.abstract.number :as an]
@@ -55,7 +55,9 @@
             [sicmutils.value :as v]
             [sicmutils.matrix :as matrix]
             [sicmutils.series :as series]
-            [sicmutils.util :as u #?@(:cljs [:refer-macros [import-vars]])]
+            [sicmutils.util :as u
+             #?@(:cljs [:refer [import-def import-vars]
+                        :include-macros true])]
             [sicmutils.util.aggregate]
             [sicmutils.util.stream :as us]
             [sicmutils.numerical.derivative]
@@ -112,7 +114,7 @@
 (defmacro using-coordinates [& args]
   `(cc/using-coordinates ~@args))
 
-(def print-expression simp/print-expression)
+(import-def simp/print-expression)
 
 (defn ref
   "A shim so that ref can act like nth in SICM contexts, as clojure core ref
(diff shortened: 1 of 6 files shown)


src/sicmutils/env.cljc
```diff
@@ -153,27 +155,30 @@
...
+(def ^{:doc "The mathematical constant [Pi](https://en.wikipedia.org/wiki/Pi)."}
+  pi Math/PI)
+(def ^{:doc "The negation of the mathematical
+constant [Pi](https://en.wikipedia.org/wiki/Pi)."}
+  -pi (g/- Math/PI))
 
-(def s:generate structure/generate)
-(def m:generate matrix/generate)
-(def v:make-basis-unit structure/basis-unit)
-(def qp-submatrix #(matrix/without % 0 0))
```
sritchie, line 162: I moved this into `hamiltonian` and used it in the one place that GJS uses it too, then aliased it back in.

src/sicmutils/util.cljc
```diff
@@ -44,15 +44,16 @@
...
-  "import a single fn or var:
+(declare illegal)
 
-  ```clojure
-  (import-def a b) => (def b a/b)
-  ```"
-  [from-ns def-name]
-  (let [from-sym# (symbol (str from-ns) (str def-name))]
-    `(def ~def-name ~from-sym#)))
+(defmacro import-def
```
sritchie, line 49: This and the version below exist to make Potemkin-like stuff work in Clojurescript... but in CLJS, `resolve` is a special form, so we can't actually go get a `var` from a symbol inside of a macro.

I also can't just pass a var directly, since `#'inc` resolves as `(var 'inc)`; I need the actual var OUTSIDE of the final-emitted form, so I can figure out what name to use.

@pangloss, any ideas here?

src/sicmutils/util.cljc
```diff
@@ -69,12 +70,13 @@
     ...
    (def d m.n.ns2/d)
     ... etc
-  ```
-  "
+  ```"
   [& imports]
   (let [expanded-imports (for [[from-ns & defs] imports
-                               d defs]
-                           `(import-def ~from-ns ~d))]
```
sritchie, line 77: my old `import-def` wasn't used anywhere else and didn't mirror potemkin syntax properly.

src/sicmutils/value.cljc
```diff
@@ -31,11 +31,9 @@
                            compare core-compare}
                   #?@(:cljs [:exclude [zero? number? = compare]]))
   (:require [sicmutils.util :as u]
-            #?(:clj [potemkin :refer [import-def]])
```
sritchie, line 34: random import!

src/sicmutils/util.cljc
pangloss: I looked at this but unfortunately don't have any suggestions. It gets very confusing with the macros being regular Clojure and being strictly separated from the Clojurescript runtime.

sritchie: It's more confusing even than that, since in bootstrapped Clojurescript, the cljs compiler CAN eat the macros.

The reason potemkin style imports can't work in Clojurescript turns out to be that even if you COULD do the resolve on the old var and alter the metadata of the new var, the vars get redefined internally and smash any mutations. No problem... maybe there is a better fix, but at least this gets the docs looking better!

PR #282, opened 2021-02-15
Freeze vector into 'up' vs a reader literal

sritchie:
From the CHANGELOG:

- #282 modifies the `sicmutils.value/freeze` implementation for Clojure vector
  to freeze vectors into the same representation as an `up` structure. This
  makes rendering these forms much more simple and matches the `scmutils`
  behavior.

- `sicmutils.structure.Structure` implements `clojure.lang.{Indexed, IReduce}`
  on the JVM, allowing it to act more like a vector (#282). (The CLJS
  implementation already did this.) `(vec (up 1 2 3))` now works correctly.


sritchie:
@littleredcomputer this was a bit involved in terms of changes to the tests. This came up while porting my SICM exercises; for the sake of the TeX renderer, and for our claims about `vector` and `up` being identical, I think this is the right way to go.

Thoughts?

Issue #151, opened 2020-11-06
ch6 tests behave differently depending on seemingly equivalent series g/-


sritchie:
Here's a reproduction. You can make this test fail by redefining the existing `seq:-` in series.cljc:

```clojure
(defn- seq:- [f g]
  (seq:+ f (seq:negate g)))
```

to:

```
(defn- seq:- [f g]
  (map g/- f g))
```

Which should be equivalent. Here's the check:

```clojure
(deftest ch6-repro
  "Evaluate this in sicmutils.sicm.ch6_test.cljc. If you defined g/- for series
  as `(map g/- f g)`, this test will return a different result than if you
  define it as `(g/+ f (g/negate g))`."
  (let [L (e/Lie-derivative
           (fn [[_ theta ptheta]]
             (/ (* -1 'a 'b (sin theta))
                ptheta)))
        H (e/series
           (fn [[_ _ ptheta]]
             (/ (square ptheta)
                (* 2 'a)))
           (fn [[_ theta _]]
             (* -1 'e 'b (cos theta))))
        E (((exp (* 'e L)) H) (up 't 'theta 'p_theta))]
    (is (= '((/ (expt p_theta 2) (* 2 a))
             0
             (/ (* a (expt b 2) (expt e 2) (expt (sin theta) 2))
                (* 2 (expt p_theta 2))))
           (take 3 (simplify E))))))
```

With the existing definition, the 3rd sequence entry is

```clojure
(/ (* a (expt b 2) (expt e 2) (expt (sin theta) 2)) (* 2 (expt p_theta 2)))
```

If you use the new `g/-` based definition, you pick up a factor of 3 in the third term:

```clojure
(/ (* 3N a (expt b 2) (expt e 2) (expt (sin theta) 2)) (* 2 (expt p_theta 2)))
```

Seems like a bug with the simplifier.

sritchie:
@littleredcomputer , does this ring any bells?

littleredcomputer:
It does not. This needs investigation, which I can do starting 11/9, given
other commitments.

The details of the simplifier is another field I want to look at. Doing
some quickie, `printf` style instrumentation, I found that _a lot_ of time
is spent, for example, canonicalizing the order of mixed partial
derivatives, essentially, bubble-sorting them by slot number. You can't
really partially differentiate something that many times in the Universe in
which we're living, so the O(n^2) isn't really a problem, but rather the
fact that we keep checking for it.

Made me think a good optimization would be just to handle that at
[...]

PR #285, opened 2021-02-15
Fix simplifier bug, seq:- more straightforward

sritchie:
From the CHANGELOG:

- #285 fixes a bug that prevented `sin / cos` from simplifying into a `tan` in
  the numerator, and makes `seq:-` slightly more efficient (closing heisenbug
  #151).


test/sicmutils/function_test.cljc
```diff
@@ -528,8 +528,8 @@
                               (gen/tuple
                                (gen/vector sg/real n)
                                (gen/vector sg/real n)))]
-            (is (= (apply (apply f/arg-scale g/+ factors) args)
-                   (apply g/+ (map g/* factors args)))))
+            (is (== (apply (apply f/arg-scale g/+ factors) args)
```
sritchie, line 531: fixes a flaky test.

PR #288, opened 2021-02-17
Instrumentation Ideas

pangloss:
While working on this project I found it difficult to be sure which function was actually getting dispatched with given arguments, and without searching the source code it was not easy to know what is implemented for a given function. This is an attempt to provide some tooling to address this.

The basic tools are `g/method-dispatched` and an added keyword argument to any function declared with `g/defgeneric`: `:specifics`:

```clojure
(g/method-dispatched g/invert 1N)
;; => [:sicmutils.value/real]

(g/invert :specifics)
;; => ([:sicmutils.series/series]
;;     [:sicmutils.value/function]
;;     [:sicmutils.differential/differential]
;;     [:sicmutils.structure/structure]
;;     [:sicmutils.complex/complex]
;;     [:sicmutils.expression/numeric]
;;     [clojure.lang.Keyword]
;;     [:sicmutils.value/real]
;;     [:sicmutils.series/power-series]
;;     [:sicmutils.matrix/matrix]
;;     [:sicmutils.polynomial/polynomial])
```

The other idea is to optionally extend `defmethod` to capture data about nested calls to operators. This just provides a mechanism that the user can then work with in the repl as needed. Here is an example:

```clojure
(def calls (atom []))
(defn call-stats [call]
  (swap! calls conj call))

(defn print-call-info [calls]
  (doseq [{:keys [name dispatch-with dispatch args chain result]} calls]
    (do
      (println (apply str (repeat (count chain) "  "))
               (if result
                 (str "=> " (pr-str result))
                 (pr-str (list* name args))))
      (when chain
        (println (apply str (repeat (inc (count chain)) "  "))
                 "types " dispatch-with)
        (println (apply str (repeat (inc (count chain)) "  "))
                 "using " dispatch)))))


(binding [sicmutils.generic/*generic-call-tap?* (constantly true)
          sicmutils.generic/*on-generic-call* call-stats]
  (((partial 2) (L3-central 'm V)) spherical-state))

(print-call-info @calls)
```

producing

```
(- #function[sicmutils.demo/T3-spherical/fn--48533] #function[sicmutils.demo/L3-central/Vs--48546])
   types  [:sicmutils.value/function :sicmutils.value/function]
   using  [:sicmutils.value/function :sicmutils.value/function]
 => #function[clojure.lang.AFunction/1]
 (partial-derivative #function[clojure.lang.AFunction/1] (2))
   types  [:sicmutils.value/function clojure.lang.ArraySeq]
   using  [:sicmutils.value/function clojure.lang.Sequential]
 => #function[clojure.lang.AFunction/1]
 (square D[[] → rdot [1004] → 1])
   types  [:sicmutils.differential/differential]
   using  [:sicmutils.differential/differential]
   (* rdot rdot)
     types  [clojure.lang.Symbol clojure.lang.Symbol]
     using  [:sicmutils.expression/numeric :sicmutils.expression/numeric]
   => (* rdot rdot)
   (+ rdot rdot)
     types  [clojure.lang.Symbol clojure.lang.Symbol]
     using  [:sicmutils.expression/numeric :sicmutils.expression/numeric]
   => (+ rdot rdot)
 => D[[] → (* rdot rdot) [1004] → (+ rdot rdot)]
...
```

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -12,4 +12,5 @@
         nrepl/nrepl {:mvn/version "0.7.0"},
         cljsjs/complex {:mvn/version "2.0.11-0"},
         borkdude/sci {:mvn/version "0.1.1-alpha.8"},
-        cljsjs/bigfraction {:mvn/version "4.0.12-0"}}}
+        cljsjs/bigfraction {:mvn/version "4.0.12-0"}}
+ :aliases {:repl {:jvm-opts ["-Dsicmutils.debug_generics=true"]}}}
--- a/src/sicmutils/abstract/function.cljc
+++ b/src/sicmutils/abstract/function.cljc
@@ -340,5 +340,5 @@
 ;; We can install one more method - [[sicmutils.generic/simplify]] simplifies
 ;; the attached name, but does not return its own function.
 
-(defmethod g/simplify [Function] [a]
+(g/defmethod g/simplify [Function] [a]
   (g/simplify (name a)))
--- a/src/sicmutils/abstract/number.cljc
+++ b/src/sicmutils/abstract/number.cljc
@@ -141,9 +141,9 @@
 
 (defn- defunary [generic-op op-sym]
   (if-let [op (sym/symbolic-operator op-sym)]
-    (defmethod generic-op [::x/numeric] [a]
+    (g/defmethod generic-op [::x/numeric] [a]
       (literal-number (op (numerical-expression a))))
-    (defmethod generic-op [::x/numeric] [a]
+    (g/defmethod generic-op [::x/numeric] [a]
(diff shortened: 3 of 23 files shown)


sritchie:
Adding https://github.com/gfredericks/debug-repl here as a note, since you had mentioned this as a great resource. @pangloss , thank you for this! It's in my queue, which I know I keep mentioning.

pangloss:
The original version of this PR became a big mess because I had replaced all of the calls to `defmethod` with `g/defmethod`. That aged faster than a fruitfly. I've cleaned that up in the hope that the core idea here can be a little more readily understandable now.

PR #289, opened 2021-02-17
Add SCI namespaces, add relaxation on arity checks

sritchie:
From the CHANGELOG:

- #289 adds many namespaces to `sicmutils.env.sci`:

  - `sicmutils.{complex,expression,modint,numsymb,polynomial,ratio,rational-function,util,value}`
  - `sicmutils.abstract.number`
  - `sicmutils.expression.analyze`
  - `sicmutils.numerical.elliptic`
  - `sicmutils.util.{aggregate,stream}`

  - #289 also introduces `sicmutils.function/*strict-arity-checks*` to allow the
    user to toggle whether or not to throw exceptions if the system thinks that
    arities are incompatible. It turns out that inside of an SCI environment,
    the usual tricks for detecting arities fail, causing errors in many
    expressions. To get around this, `*strict-arity-checks*` is FALSE by
    default.


src/sicmutils/function.cljc
```diff
@@ -82,8 +82,7 @@
    (let [new-meta (-> (meta f)
                       (merge m)
                       (assoc :arity arity))]
-     (if new-meta
-       (with-meta f new-meta)))))
+     (with-meta f new-meta))))
```
sritchie, line 85: I don't think we'd ever hit the `else` branch, but what was this!

src/sicmutils/function.cljc
```diff
@@ -240,7 +239,7 @@
 ;;   [:at-least m]
 
 #?(:clj
-   (do (defn- arity-map [f]
+   (do (defn ^:no-doc arity-map [f]
```
sritchie, line 242: this is better than private, so they're at least available to experiment with...

src/sicmutils/function.cljc
```diff
@@ -296,27 +295,29 @@
 
    :cljs
    (do
-     (defn- variadic?
+     (defn ^:no-doc variadic?
        "Returns true if the supplied function is variadic, false otherwise."
        [f]
-       (boolean (.-cljs$lang$maxFixedArity f)))
+       (boolean
+        (.-cljs$core$IFn$_invoke$arity$variadic f)))
```
sritchie, line 302: this is more correct. The previous impl was failing for `defn`-created functions.

src/sicmutils/function.cljc
```diff
@@ -342,14 +343,21 @@
...
   definitely worth memoizing."}
   reflect-on-arity
   (memoize
    #?(:cljs js-arity :clj jvm-arity)))
 
+(def ^{:dynamic true
+       :doc "If true, attempting to pass two functions of incompatible arity
+  into any binary function, or into [[combine-arities]], will throw. False by
+  default."}
+  *strict-arity-checks*
```
sritchie, line 358: @littleredcomputer , here is a fairly large change. I'm introducing this variable and setting it to false because the SCI / nextjournal environment was choking on many valid expressions.

It turns out that inside of an SCI environment, the usual tricks for detecting arities via js object inspection fail... so, I think `false` and permissive by default is the right move. Thoughts?

src/sicmutils/function.cljc
```diff
@@ -456,32 +468,9 @@
...
-                    [:exactly 8]
-                    #(op (f1 %1 %2 %3 %4 %5 %6 %7 %8) (g1 %1 %2 %3 %4 %5 %6 %7 %8))
-                    [:exactly 9]
-                    #(op (f1 %1 %2 %3 %4 %5 %6 %7 %8 %9) (g1 %1 %2 %3 %4 %5 %6 %7 %8 %9))
-                    [:exactly 10]
-                    #(op (f1 %1 %2 %3 %4 %5 %6 %7 %8 %9 %10) (g1 %1 %2 %3 %4 %5 %6 %7 %8 %9 %10))
-                    [:at-least 0]
-                    #(op (apply f1 %&) (apply g1 %&))
-                    (u/illegal (str  "unsupported arity for function arithmetic " arity)))
+              (-> (fn [& args]
```
sritchie, line 471: Clojure can go up to arities of 21; rather than add those here I figured I'd do the Clojure Thing and trust that if `joint-arities` says we're okay to proceed we don't have to do any further checking.

PR #295, opened 2021-02-19
rendering impls for floor, ceil, integer-part, fractional-part, modulo, remainder

sritchie:
This is a follow-up to #278. Here's a test function:

```clojure
(println
 (->infix
  (g/floor (g// (g/floor (g/modulo 'n 'x)) 'n))))
```

Demo for some riffs on that:

image

## Tasks

- [x] TeX Rendering! Notes on [fractional-part](https://math.stackexchange.com/questions/1882055/notation-for-specifying-the-fractional-part-of-a-number), and [integer-part](https://math.stackexchange.com/questions/1882055/notation-for-specifying-the-fractional-part-of-a-number), [floor](https://tex.stackexchange.com/questions/118173/how-to-write-ceil-and-floor-in-latex), [ceiling](https://tex.stackexchange.com/questions/118173/how-to-write-ceil-and-floor-in-latex).
- [x] change rendering to `int` and `frac` for integer and fractional parts
- [x] `->Javascript` and compile have to handle
  - [x] remainder
  - [x] modulo,
  - [x] integer-part
  - [x] fractional-part
- [x] switch to [Math.trunc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc) for CLJS integer-part impl
- [x] CHANGELOG entry
- [x] We'll need to look up the unicode characters for the infix renderer too.
- [x] Generative tests, lock in the "laws" for how these types interact
- [x] Docstrings for the new methods explaining how they relate, and the choices we've made
- [x] Change the Fraction implementation to match Clojure's Numbers.java impl

### DIFF
--- a/src/sicmutils/complex.cljc
+++ b/src/sicmutils/complex.cljc
@@ -156,9 +156,11 @@
 (defmethod g/simplify [::complex] [a] (v/freeze a))
 
 (defmethod g/integer-part [::complex] [^Complex a]
-  (complex
-   (g/integer-part (#?(:clj .getReal :cljs .-re) a))
-   (g/integer-part (#?(:clj .getImaginary :cljs .-im) a))))
+  (let [re (g/integer-part (#?(:clj .getReal :cljs .-re) a))
+        im (g/integer-part (#?(:clj .getImaginary :cljs .-im) a))]
+    (if (v/zero? im)
+      re
+      (complex re im))))
 
 (defmethod g/fractional-part [::complex] [^Complex a]
   (complex
@@ -183,11 +185,18 @@
 #?(:clj
    (do
      (defmethod g/floor [::complex] [^Complex a]
-       (complex (g/floor (.getReal a))
-                (g/floor (.getImaginary a))))
+       (let [re (g/floor (.getReal a))
+             im (g/floor (.getImaginary a))]
+         (if (v/zero? im)
+           re
+           (complex re im))))
+
      (defmethod g/ceiling [::complex] [^Complex a]
-       (complex (g/ceiling (.getReal a))
-                (g/ceiling (.getImaginary a))))
+       (let [re (g/ceiling (.getReal a))
+             im (g/ceiling (.getImaginary a))]
+         (if (v/zero? im)
+           re
+           (complex re im))))
 
      (defmethod g/sub [::complex ::complex] [^Complex a ^Complex b] (.subtract a b))
      (defmethod g/sub [::complex ::v/real] [^Complex a n] (.subtract a (double n)))
--- a/src/sicmutils/expression/compile.cljc
+++ b/src/sicmutils/expression/compile.cljc
@@ -102,6 +102,9 @@
    'cosh #(Math/cosh %)
    'sinh #(Math/sinh %)
    'tanh #(Math/tanh %)
+   'floor #(Math/floor %)
+   'ceiling #(Math/ceil %)
+   ;; TODO: modulo, remainder
    #?@(:cljs
        ;; JS-only entries.
        ['acosh #(Math/acosh %)
--- a/src/sicmutils/expression/compile.cljc
+++ b/src/sicmutils/expression/compile.cljc
@@ -90,7 +90,7 @@
    '/ /
    'expt #(Math/pow %1 %2)
    'sqrt #(Math/sqrt %)
-   'abs (fn [^double x] (Math/abs x))
+   'abs #(Math/abs %)
    'log #(Math/log %)
    'exp #(Math/exp %)
    'cos #(Math/cos %)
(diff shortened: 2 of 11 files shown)


sritchie:
@pangloss , giving a go at some of the remaining stuff here.

pangloss:
I noticed in my travels that possible representations for integer-part and fractional-part are `[ x ]` and `{ x }`. That seems like a recipe for confusion. Fortunately the alternative `int( x )` and `frac( x )` seem much safer. WA uses the latter.

sritchie:
@pangloss yup, I added the confusing versions here to verify that they are, in fact, very confusing! I like your suggestion.

src/sicmutils/complex.cljc
```diff
@@ -156,9 +156,11 @@
 (defmethod g/simplify [::complex] [a] (v/freeze a))
 
 (defmethod g/integer-part [::complex] [^Complex a]
-  (complex
-   (g/integer-part (#?(:clj .getReal :cljs .-re) a))
-   (g/integer-part (#?(:clj .getImaginary :cljs .-im) a))))
+  (let [re (g/integer-part (#?(:clj .getReal :cljs .-re) a))
+        im (g/integer-part (#?(:clj .getImaginary :cljs .-im) a))]
+    (if (v/zero? im)
```
sritchie, line 161: Three of the new functions give us a case where we have an integer in hand and can trust our `zero?` result. If zero, return just the real part.

src/sicmutils/expression/compile.cljc
```diff
@@ -90,7 +90,7 @@
    '/ /
    'expt #(Math/pow %1 %2)
    'sqrt #(Math/sqrt %)
-   'abs (fn [^double x] (Math/abs x))
```
sritchie, line 93: this type hint wasn't doing anything... and if this DOES make a difference I'd change it everywhere.

src/sicmutils/expression/compile.cljc
```diff
@@ -102,6 +102,15 @@
...
    'tanh #(Math/tanh %)
+   'floor #(Math/floor %)
+   'ceiling #(Math/ceil %)
+   'modulo mod
+   'remainder rem
+   'quotient quot
+   'integer-part #?(:clj long
+                    :cljs #(Math/trunc %))
+   'fractional-part (fn [^double x]
+                      (- x (Math/floor x)))
```
sritchie, line 113: I'll make a ticket to get the simplifier to do the transform for us; after that we can remove this special-casing.

src/sicmutils/ratio.cljc
```diff
@@ -249,12 +242,17 @@
          (g/div (g/sqrt (u/double (numerator a)))
                 (g/sqrt (u/double (denominator a))))))
 
+     (defmethod g/modulo [Fraction Fraction] [^Fraction a ^Fraction b]
```
sritchie, line 245: I added type hints, `promote`, and let the default values below handle the non-fraction-to-fraction cases.

test/sicmutils/collection_test.cljc
```diff
@@ -33,7 +33,7 @@
 
 (deftest vector-tests
   (testing "Vector protocol implementations"
-    (checking "f/arity" 100 [v (gen/vector sg/integer)]
+    (checking "f/arity" 100 [v (gen/vector sg/any-integral)]
```
sritchie, line 36: I realized I had some of these tests just using `goog.math.Integer`, vs the whole integral generator!

test/sicmutils/numbers_test.cljc
```diff
@@ -35,6 +35,18 @@
 
 (def near (v/within 1e-12))
 
+(defn gen-integer
```
sritchie, line 38: I need to make the originals take max sizes...

pangloss (approved):
This looks great.

I thought the removal of the various specializations of remainder and modulo for Fraction would cause those methods to return reals, but it looks like the upcast-number technique which I hadn't previously noticed covers it since testing that, the results are correct.

Just one very minor nit on the gen-real with max arg.

Thanks for putting the polish on this!

test/sicmutils/numbers_test.cljc
```diff
@@ -35,6 +35,18 @@
...
+  "Integral generator that takes a `max` magnitude size parameter."
+  ([] sg/any-integral)
+  ([max] (gen/fmap #(g/modulo % (int max))
+                   sg/any-integral)))
+
+(defn gen-real
+  "Integral generator that takes a `max` magnitude size parameter."
+  ([] sg/real)
+  ([max] (gen/fmap #(g/modulo % (int max))
```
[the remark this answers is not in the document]
sritchie: Generative testing revealed incorrect results like these for (rem bigint float):

```clojure
(rem 9037601485536036300227801N 1e4)
```

double on the right gets this wrong, and returns a number greater than the modulus. I coerced so I could still write 1e4. Fishy stuff!

PR #296, opened 2021-02-20
Exercise 1.44 benchmark, stress test, plots [in progress]

sritchie:
GJS sent me some code and examples that I'm porting as a basis for plots. Coming soon!

- [ ] TODO this is actually a good test because it exposed some sadness with the simplifier. Convert this over from the Clerk notebook and at least get the non-chaotic test in. Forget the other stuff.

PR #297, opened 2021-02-22
structural division fix, series simplify fix

sritchie:
These issues came up while working on #296.

 From the CHANGELOG:

- `g/simplify` on `Series` or `PowerSeries` now keeps the type intact; before
  #297, simplifying a series returned a bare Clojure sequence.

- division between two `Structure` instances `a` and `b` now (as of #297)
  returns a new structure instance `(/ a b)` that matches the contract `(= a (*
  b (/ a b)))`. Previously, the division would not necessarily contract with `b`
  to return `a`, resulting in problems with the double pendulum exercise of
  #296.


src/sicmutils/matrix.cljc
```diff
@@ -455,11 +455,10 @@
   ([m outer-orientation inner-orientation t?]
    {:pre [(s/valid-orientation? outer-orientation)
           (s/valid-orientation? inner-orientation)]}
-   (let [m'   (if t? (transpose m) m)
-         rows (matrix->vector m')]
+   (let [m'   (if t? (transpose m) m)]
      (s/make outer-orientation
              (mapv #(s/make inner-orientation %)
-                   rows)))))
+                   m')))))
```
sritchie, line 461: matrices already act like sequences (and associative things with random access), so there's no need to try and unpack their contents explicitly.

src/sicmutils/matrix.cljc
```diff
@@ -636,7 +635,7 @@
   other type is supplied."
   [m]
   {:pre [(row? m)]}
-  (first m))
+  (nth m 0))
```
sritchie, line 638: faster!

src/sicmutils/matrix.cljc
```diff
@@ -875,6 +873,11 @@
       (s/opposite a' (map #(s/opposite a' %) a'))
       a')))
 
+(defmethod g/div [::s/structure ::s/structure] [rv s]
```
sritchie, line 876: this was killing me... lots to learn here but this definitely Does the Right Thing if you have nested coordinate entries in the local tuple.

src/sicmutils/numerical/ode.cljc
```diff
@@ -166,29 +166,32 @@
   #?(:cljs
      (let [total-time (us/stopwatch :started? false)
            latest     (atom 0)]
-       (fn [initial-state step-size t {:keys [observe] :as opts}]
```
sritchie, line 169: This was missing an arity that didn't need an options map supplied. The CLJ version did have this new arity!

src/sicmutils/series.cljc
```diff
@@ -729,8 +729,7 @@
     (ctor (i/sqrt (seq s))))
 
   (defmethod g/simplify [kind] [s]
-    (map g/simplify (seq s))))
-
+    (fmap g/simplify s)))
```
sritchie, line 732: the other main fix.

Issue #255, opened 2021-02-03
`g/simplify` on Literal, Structure should return Literal, Structure


sritchie:
Right now, we get back a bare number or expression. (Also a problem for structures!)

The reason this will be tough is that, right now, we expect expressions to compare to sequences using `(= '(* x 2) (g/* 'x 2))`. This will only work if `Literal` implements `Sequential` and `Seqable`... but if you do that to, say, `(literal-number 'x)`, a `Literal` that contains a NON-sequence, you see this:

```clojure
sicmutils.abstract.number> (= [] (literal-number 10))
Execution error (IllegalArgumentException) at sicmutils.expression.Literal/seq (expression.cljc:90).
Don't know how to create ISeq from: java.lang.Long
```

There is no way to conditionally implement these two interfaces, unfortunately.

The solution will be to implement two types, one for sequence-like expressions and the other for the "atom" cases of `(literal-number 'x)` and `(literal-number 12)`.

I think we really do want to be able to have the exactness-preserving `(literal-number 12)` behavior, so we need a wrapper type.

sritchie:
Okay! I solved this by making `v/=` do all the work that `=` was able to do before. Huzzah!

PR #298, opened 2021-02-22
Prep for 0.16.0 release

sritchie:
In addition to the CHANGELOG entries I'll add below, this PR:

- derivative implementations for floor, ceiling, integer-part and fractional-part
- consolidates the arithmetic operators like `+` and their binary implementations into one multi-arity function for each of `+`, `-`, `*` and `/`, and adds solid docs to each of these.
- `sicmutils.util.def` hosts `defgeneric` AND the Potemkin-like var importing code that `sicmutils.env` uses.
- adds some explanation to the top of `sicmutils.generic` about what generic is all about.

From the CHANGELOG:

- In JVM Clojure (as of #298), `sicmutils.expression.compile` defaults to
   `clojure.core/eval` to compile functions, while Clojurescript defaults to
   [SCI](https://github.com/borkdude/sci). The performance is much faster for
   numerical routines and worth the slightly different default behavior.

   To use to SCI compilation on the JVM, wrap your form in a binding:

   ```clojure
   (require '[sicmutils.expression.compile :as compile])

   (binding [compile/*mode* :sci]
     (my-compiler-triggering-function))
   ```

   To set the mode permanently, use `compile/set-compiler-mode!`:

   ```clojure
   (compile/set-compiler-mode! :sci)

   (my-compiler-triggering-function)
   ```

   The options allowed as of `0.16.0` are `:sci` and `:native`.

 - `clojure.lang.Var` implements the `sicmutils.value/Value` protocol, allowing
   it to respond appropriately with its name to `v/freeze` (#298).

 - `sicmutils.modint/make` now verifies with a precondition that its two
   arguments are both `v/integral?` (#298). We need this constraint now that
   `g/modulo` is defined for more types.

 - `g/simplify` called with an argument `x` of type `Series`, `PowerSeries`,
   `Matrix`, `Operator`, `Complex` and `sicmutils.abstract.function/Function` now
   return an instance of type `x`, performing appropriate simplifications if
   possible. before #297 and #298, these operation would return bare symbols or
   sequences. A future release will make this change for `Structure` and
   `Literal` too, once #255 is resolved.


src/sicmutils/collection.cljc
```diff
@@ -38,6 +39,9 @@
 ;; act as `zero?`, but they can't act as `one?` or `identity?`; those are
 ;; reserved for instances that have no effect on multiplication.
 
+(defmethod g/simplify [PersistentVector] [v]
```
sritchie, line 42: I moved this simplify implementations out of generic into their proper places.

src/sicmutils/complex.cljc
```diff
@@ -153,7 +153,6 @@
 (defmethod g/cosh [::complex] [^Complex a] (.cosh a))
 (defmethod g/sinh [::complex] [^Complex a] (.sinh a))
 (defmethod g/tanh [::complex] [^Complex a] (.tanh a))
-(defmethod g/simplify [::complex] [a] (v/freeze a))
```
sritchie, line 156: We're moving away from this, so let the default take over.

src/sicmutils/differential.cljc
```diff
@@ -1245,6 +1246,35 @@
...
+(defn- discont-at-integers [f dfdx]
+  (let [f (lift-1 f (fn [_] dfdx))
+        f-name (v/freeze f)]
+    (fn [x]
+      (if (v/integral? (finite-term x))
+        (u/illegal
+         (str "Derivative of g/" f-name " undefined at non-integer points."))
+        (f x)))))
+
+(defunary g/floor
```
sritchie, line 1258: @pangloss , here are the derivatives for these functions.

src/sicmutils/generic.cljc
```diff
@@ -25,125 +25,173 @@
...
-  (:refer-clojure :rename {mod core-mod}
-                  :exclude [/ + - * divide #?(:cljs mod)])
+  (:refer-clojure :exclude [/ + - * divide])
   (:require [sicmutils.value :as v]
-            [sicmutils.util :as u])
-  #?(:cljs (:require-macros [sicmutils.generic :refer [defgeneric]]))
-  #?(:clj
-     (:import [clojure.lang Keyword LazySeq PersistentVector Symbol Var])))
-
-(defmacro ^:private fork
```
sritchie, line 36: I moved these out to tidy things up and started the process of documenting this appropriately.

src/sicmutils/matrix.cljc
```diff
@@ -851,7 +851,7 @@
 (defmethod g/tanh [::square-matrix] [m] (series/tanh-series m))
 (defmethod g/asinh [::square-matrix] [m] (series/asinh-series m))
 (defmethod g/atanh [::square-matrix] [m] (series/atanh-series m))
-(defmethod g/simplify [::matrix] [m] (->> m (fmap g/simplify) v/freeze))
```
sritchie, line 854: no more freeze!

src/sicmutils/operator.cljc
```diff
@@ -426,5 +426,3 @@
 
 (defmethod g/square [::operator] [o] (o:* o o))
 (defmethod g/cube [::operator] [o] (o:* o (o:* o o)))
-
-(defmethod g/simplify [::operator] [o] (name o))
```
sritchie, line 430: I didn't implement this as `(simplify (name o))` because the expression here does NOT simplify as if it's a real. There are different operator simplification rules.

src/sicmutils/value.cljc
```diff
@@ -210,8 +211,20 @@
   (one-like [o] (u/unsupported "nil doesn't support one-like."))
   (identity-like [o] (u/unsupported "nil doesn't support identity-like."))
   (freeze [_] nil)
+  (exact? [x] false)
   (kind [_] nil)
 
+  Var
+  (zero? [_] false)
```
sritchie, line 218: this allows `v/freeze` to work properly here.

src/sicmutils/differential.cljc
```diff
@@ -1245,6 +1246,35 @@
...
+(defunary g/fractional-part
+  (discont-at-integers g/fractional-part 1))
+
+(defunary g/floor
+  (let [f (lift-1 g/floor (fn [_] 0))]
+    (fn [x]
+      (if (v/integral? (finite-term x))
+        (u/illegal
+         (str "Derivative of g/floor undefined at non-integer points."))
+        (f x)))))
```
pangloss, line 1276: I'm not sure about this, but the message in the exception seems to indicate that it belongs in the else position, rather than the then position?

pangloss: The call to `str` is also redundant here

sritchie: Good catch, it was actually the full function definition that was redundant, given my version above!

PR #304, opened 2021-02-28
return Literal and Structure on simplify, not bare sequences

sritchie:
This is a big one!! From the CHANGELOG:

   - aliases `sicmutils.operator/anticommutator` and `sicmutils.util/bigint?`
     into `sicmutils.env`

   - implements `v/=` properly for sequences, `Differential`, `Complex`,
     `Structure` and `Matrix` instances

   - in `sicmutils.env`, `v/=` now overrides `clojure.core/=`. `v/=` should act
     identically to `clojure.core/=` everywhere; the difference is that its
     behavior is customizable, so we can make `Differential` instances equal to
     numbers, or complex numbers with a 0 imaginary part equal to real numbers
     with the same real part.

     `v/=` may not drop recursively down into, say, Clojure maps. Please open an
     issue if you find a case like this!

   - BIG CHANGE: `Literal` and `Structure` instances now KEEP their type under
     `g/simplify`. This means that you can no longer make comparisons like this:

 ```clojure
 ;; this worked before, and was used all over the tests (probably not in much
 ;; user code!)
 (clojure.core/= '(* 3 x)
                 (simplify (+ 'x 'x 'x)))
 ;;=> false
 ```

   Instead, use `v/=` (which is now aliased into `sicmutils.env`):

 ```clojure
 ;; `v/=` will do the right thing by unwrapping the literal expression on the
 ;; right:
 (v/= '(+ x y) (+ 'x 'y))
 ;;=> true
 ```


src/sicmutils/abstract/number.cljc
```diff
@@ -114,18 +114,14 @@
...
-       (= (x/expression-of l) n)))
-
 ;; This installs equality into `v/=` between symbolic expressions (and symbols,
-;; see inheritance above) and anything in the standard numeric tower.
+;; see inheritance above), sequences where appropriate, and anything in the
+;; standard numeric tower.
 
-(defmethod v/= [::x/numeric ::v/number] [l r] (literal=num l r))
-(defmethod v/= [::v/number ::x/numeric] [l r] (literal=num r l))
+(defmethod v/= [::x/numeric v/seqtype] [l r] (v/= (x/expression-of l) r))
```
sritchie, line 121: these weren't correctly comparing to sequences before; now `v/=` should act JUST like everything did before the `freeze` removal.

src/sicmutils/differential.cljc
```diff
@@ -1221,6 +1221,12 @@
 ;; provides [[defunary]] and [[defbinary]] calls for all of the generic
 ;; operations for which we know how to declare partial derivatives.
 
+;; First, install `equiv` as to perform proper equality between `Differential`
+;; instances and scalars. `equiv` compares on only the finite part, not the
+;; differential parts.
+
+(defbinary v/= equiv)
```
sritchie, line 1228: phew, that was easy!

src/sicmutils/expression/compile.cljc
```diff
@@ -492,7 +493,8 @@
         generic-state  (struct/mapr (fn [_] (gensym 'y)) initial-state)
         g              (apply f generic-params)
         body           (cse-form
-                        (g/simplify (g generic-state)))
+                        (v/freeze
```
sritchie, line 496: now that we don't automatically freeze, we have to do this manually.

src/sicmutils/expression/render.cljc
```diff
@@ -619,7 +619,8 @@
     (fn [x & {:keys [symbol-generator parameter-order deterministic?]
              :or {symbol-generator (make-symbol-generator "_")
                   parameter-order sort}}]
-      (let [params (set/difference (x/variables-in x) operators-known)
+      (let [x      (v/freeze x)
```
sritchie, line 622: same here, on the way in to the JS renderer.

src/sicmutils/matrix.cljc
```diff
@@ -230,7 +230,7 @@
        (let [^Matrix m that]
          (and (= (.-r this) (.-r m))
               (= (.-c this) (.-c m))
-              (= (.-v this) (.-v m))))))
+              (v/= (.-v this) (.-v m))))))
```
sritchie, line 233: this way it will drop to `v/=` recursively.

src/sicmutils/value.cljc
```diff
@@ -254,6 +254,10 @@
   #?(:clj  (== l r)
      :cljs (identical? l r)))
 
+(defmethod = [seqtype seqtype] [l r]
+  (and (= (count l) (count r))
+       (every? true? (map = l r))))
```
sritchie, line 259: recursively drops down using `v/=`

CHANGELOG.md
```diff
@@ -2,6 +2,43 @@
...
+                (simplify (+ 'x 'x 'x)))
+;;=> false
+```
+
+  Instead, use `v/=` (which is now aliased into `sicmutils.env`):
+
+```clojure
+;; `v/=` will do the right thing by unwrapping the literal expression on the
+;; right:
+(v/= '(+ x y) (+ 'x 'y))
```
littleredcomputer, line 38: That is bold. I hadn't thought of that. So an S expression is equivalent to a sicmutils expressing holding an S expression that would have been =. I can deal with that. Does it work as well to use an expression constructor on the left?

src/sicmutils/abstract/number.cljc
```diff
@@ -114,18 +114,18 @@
...
-
 ;; This installs equality into `v/=` between symbolic expressions (and symbols,
-;; see inheritance above) and anything in the standard numeric tower.
-
-(defmethod v/= [::x/numeric ::v/number] [l r] (literal=num l r))
-(defmethod v/= [::v/number ::x/numeric] [l r] (literal=num r l))
+;; see inheritance above), sequences where appropriate, and anything in the
+;; standard numeric tower.
+
+(defmethod v/= [Symbol v/seqtype] [_ _] false)
```
littleredcomputer, line 121: Is there a way to say that v/= is universally false unless a real rule asserts otherwise?

src/sicmutils/differential.cljc
littleredcomputer: impressive.

src/sicmutils/structure.cljc
```diff
@@ -320,6 +320,17 @@
   #(get-in % indices))
 
 ;; ## Structure Predicates
+;;
+;; `::down` instances should never be equal to collections, or `::up`. By
+;; default in Clojure, all collections compare as if they were sequences, so an
+;; up can't equal a down... but a vector would! This change fixes that.
+
+(defmethod v/= [::down ::up] [_ _] false)
```
littleredcomputer, line 328: this is another good argument for having some way of saying that v/= random things are false...

test/sicmutils/structure_test.cljc
```diff
@@ -210,6 +210,18 @@
       (is (= "(up 1 2 3)" s)))
     (is (= "(up 1 2 3)" (str (s/up 1 2 3)))))
 
+  (testing "down is not equal to up or seqs"
```
littleredcomputer, line 213: I may have overlooked them but it might be nice to have some tests that assert that that v/= is deep with respect to structures by proving asserting some equalities that Clojure's built-in = would say false to, like
`(is (v/= [1 2] [(complex 1 0) (complex 2 0)])` (although your refactor of constructors might make that trivial, I should fire it up and see...yes, that would be new).

littleredcomputer (approved):
very cool. So much stuff works here. The test suite is a glory to behold now.

CHANGELOG.md
sritchie: yes!

```clojure
sicmutils.env> (v/= (literal-number '(+ x y)) (+ 'x 'y))
true
```

Let me type out my thoughts here and make the best case for this, so you can swat it down if I'm leading myself down the primrose path :)

This PR had me a little queasy, since we don't want to break equality... but I am fairly sure it all works. This behavior is equivalent to what would have happened if we'd implemented `Seqable` for `Literal`, since anything `Seqable` compares by first converting each side with `seq`.

But the problem is that expressions can be either seq-like things, like `(+ 1 2)`, or they can simplify down to symbols or numbers, like `'x` or `1`.

You might say, why not just unwrap them when they get to that point? But I'm convinced that we want the ability to say `(literal-number 2)`, and have that force itself into the symbolic arithmetic tower such that `(log (literal-number 2))` returns `(log 2)`.

We use that to nice effect here: https://github.com/sicmutils/sicmutils/pull/315

src/sicmutils/abstract/number.cljc
sritchie: Yeah, it does feel like there's something that we want to express here, how any combo of distinct types that we know about  should be false.

I think we accomplish that with the default cases: https://github.com/sicmutils/sicmutils/blob/master/src/sicmutils/value.cljc#L252-L261

If you don't hit anything here, you drop back to `clojure.core/=`, which lets `equals` or `-equiv` (in the JVM or JS, respectively) handle the behavior. So `v/=` is our ultra-customizable equals function that drops back to the customizable-but-only-on-the-left-side functions that know and love.

src/sicmutils/differential.cljc
sritchie: As a reminder from previous PRs, `equiv` only compares the finite parts (and is what we've installed into `equals` for `Differential` instances), while `eq` also takes the tangents into account.

So this means that `(v/= l r)` will do the same thing for non-differential and differential arguments, which is how we want it to behave inside functions whose derivatives we're taking (to phrase it awkwardly!)

src/sicmutils/structure.cljc
sritchie: This one is interesting, because both sides are sequences. If you let these dump back to `clojure.core/=`, or the default for `seq, seq` combos, both sides get turned into sequences and their contents compared.

But UNLIKE Clojure, we know that these two sequences are very different containers indeed, and the type of container matters!

src/sicmutils/structure.cljc
```diff
@@ -320,6 +320,17 @@
...
+(defmethod v/= [::up ::down] [_ _] false)
+(defmethod v/= [::down v/seqtype] [_ _] false)
+(defmethod v/= [v/seqtype ::down] [_ _] false)
+(prefer-method v/= [::up ::down] [v/seqtype ::down])
```
sritchie, line 332: Having to write this signals to me that this double-dispatch trick might get very clunky. Let's keep our brains and eyes looking for some way to express more clearly what `v/=` is trying to do.

test/sicmutils/structure_test.cljc
sritchie: Ah, good idea! I made a ticket to get this working for maps, too (and maybe `java.util.List` and a few other cases). I added your comment there so we can add these tests as a package deal.

https://github.com/sicmutils/sicmutils/issues/316

But that example works!

```clojure
sicmutils.env> (= [1 2] [(complex 1 0) (complex 2 0)])
;;=> true
```

sritchie (commented):
Thank you for the lovely review! It's great to see your comments again, and I'm thoroughly glad that you're set up and took your time away from the little glass rectangle seriously :)

Onward!

PR #307, opened 2021-02-28
add multi-argument gcd and lcm

### LINKED ISSUE #89: Add multiple-argument gcd, lcm, patterned on `g/*` and friends
This is how the operation is defined in Scheme:

```scheme
(define (g:gcd:n args)
  (cond ((null? args) :zero)
        ((null? (cdr args)) (car args))
        (else
         (let lp
             ((as (cddr args))
              (ans (g:gcd:bin (car args) (cadr args))))
           (cond ((null? as) ans)
                 ((g:one? ans) ans)
                 (else
                  (lp (cdr as) (g:gcd:bin ans (car as)))))))))
```

The task here would be to

- rename the generics now to something like `g/-gcd` and `g/-lcm`
- expose a multi-arg `g/gcd`, `g/lcm` that wraps them and does this trick with both functions.

### LINKED ISSUE #177: Add multiple-argument gcd, lcm, patterned on `g/*` and friends
This is how the operation is defined in Scheme:

```scheme
(define (g:gcd:n args)
  (cond ((null? args) :zero)
        ((null? (cdr args)) (car args))
        (else
         (let lp
             ((as (cddr args))
              (ans (g:gcd:bin (car args) (cadr args))))
           (cond ((null? as) ans)
                 ((g:one? ans) ans)
                 (else
                  (lp (cdr as) (g:gcd:bin ans (car as)))))))))
```

The task here would be to

- `rename the generics now to something like `g/-gcd`, `g/-lcm`
- expose a multi-arg `g/gcd`, `g/lcm` that wraps them and does this trick with both functions.

kolharsam:
this PR should resolve sicmutils/placeholder#88

### DIFF
--- a/src/sicmutils/generic.cljc
+++ b/src/sicmutils/generic.cljc
@@ -447,10 +447,27 @@
 (defmethod remainder :default [n d]
   (remainder-default n d))
 
+;;; FIXME?: is the multi-argument definition required? [[gcd]], I don't fully
+;;; understand what [[add]] is doing. Where do I look?
 (defgeneric gcd 2
   "Returns the [greatest common
   divisor](https://en.wikipedia.org/wiki/Greatest_common_divisor) of the two
-  inputs `a` and `b`.")
+  inputs `a` and `b`."
+  {:name '-gcd}) ;;; FIXME?: not sure if this is right
+
+(defn gcd
+  "FIXME: add an appropiate comment"
+  [x y]
+  (if (zero? y)
+    x
+    (recur y (mod x y))))
+
+(defn -gcd
+  ([] 1)
+  ([x] x)
+  ([x y] (gcd x y))
+  ([x & rest]
+   (reduce gcd x rest)))
 
 (defgeneric lcm 2
   "Returns the [least common
--- a/src/sicmutils/generic.cljc
+++ b/src/sicmutils/generic.cljc
@@ -466,8 +466,8 @@
   ([] 1)
   ([x] x)
   ([x y] (gcd x y))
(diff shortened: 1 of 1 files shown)


kolharsam:
@sritchie I have made the changes for `gcd` alone now. I'd like to clean up the mistakes that I've made here and then based on that would like to do further work on `lcm`. I hope that is okay?

sritchie:
Hey @kolharsam ! It's probably worth doing a quick chat on Clojurians Slack if you're on that forum so we can go over what this PR is supposed to achieve! Send me a note in https://clojurians.slack.com/archives/C01ECA9AA74 via the username @sritchie and we'll discuss. Sound good?

src/sicmutils/generic.cljc
```diff
@@ -447,10 +447,27 @@
 (defmethod remainder :default [n d]
   (remainder-default n d))
 
+;;; FIXME?: is the multi-argument definition required? [[gcd]], I don't fully
+;;; understand what [[add]] is doing. Where do I look?
 (defgeneric gcd 2
```
sritchie, line 452: Ah! So the idea is that this is a generic function already, and many different types extend it. But it only works for two arguments. So, the task is:

- change this to `(defgeneric -gcd 2 ...`)
- delete the implementation you have on line `458`
- change `465` to call `-gcd` internally

Then the exposed `gcd` will be multiple arity. We'll need tests too. Let me know if that makes sense!

sritchie:
@kolharsam , closing this since it looks like it is abandoned for now. Thank you!

PR #319, opened 2021-03-05
Add symbolic derivative, boolean operations

sritchie:
I buckled and added to the simplifier for operators. Now `identity` gets stripped out of products. This will cause problems if someone names a NEW operator `identity`... but let's just hope no one does that!

From the CHANGELOG, this PR adds:

   - symbolic boolean implementations for `sym:=`, `sym:and`, `sym:or` and
     `sym:not` with infix, latex and JavaScript renderers.
   - `sym:derivative`, for purely symbolic derivatives

   The boolean operators will act just like `=`, `and` and `or` on booleans, and
   appropriately respond if just one side is a boolean. If both sides are
   symbolic, These return a form like `(= a b)`, `(and a b)` or `(or a b)`.

   The functions currently live in `sicmutils.numsymb` only; access them via
   `(numsymb/symbolic-operator <sym>)`, where `<sym>` is one of `'=`, `'and`,
   `'or`, `'not` or `'derivative`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,27 @@
 
 ## [unreleased]
 
+- #319: adds
+
+  - symbolic boolean implementations for `sym:=`, `sym:and`, `sym:or` and
+    `sym:not` with infix, latex and JavaScript renderers.
+  - `sym:derivative`, for purely symbolic derivatives
+
+  The boolean operators will act just like `=`, `and` and `or` on booleans, and
+  appropriately respond if just one side is a boolean. If both sides are
+  symbolic, These return a form like `(= a b)`, `(and a b)` or `(or a b)`.
+
+  The functions currently live in `sicmutils.numsymb` only; access them via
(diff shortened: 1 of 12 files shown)


src/sicmutils/abstract/function.cljc
```diff
@@ -34,8 +34,7 @@
             [sicmutils.polynomial]
             [sicmutils.structure :as s]
             [sicmutils.util :as u]
-            [sicmutils.value :as v]
-            [sicmutils.calculus.derivative :refer [derivative-symbol]])
```
sritchie, line 38: this nicely removes the dependency of abstract functions on derivatives, which by all rights shouldn't come before these in the dependency tree.

CHANGELOG.md
```diff
@@ -25,8 +41,8 @@ sicmutils.env> (series/exp-series D)
 sicmutils.env> (series/exp-series D)
 #object[sicmutils.series.Series
   "(+ identity
-      (* D identity)
```
sritchie, line 28: this was too good to pass up!

PR #329, opened 2021-03-19
Fix simplifier bug with exprs like (sqrt (literal-number 2))

sritchie:
From the CHANGELOG:

- #329 fixes a bug where the simplifier couldn't handle expressions like `(sqrt
  (literal-number 2))`, where literal numbers with no symbols were nested inside
  of symbolic expressions.

### DIFF
--- a/src/sicmutils/polynomial.cljc
+++ b/src/sicmutils/polynomial.cljc
@@ -29,8 +29,15 @@
             [sicmutils.value :as v]
             #?(:cljs [goog.string :refer [format]])))
 
-(def coefficient second)
-(def exponents first)
+(defn exponents [term]
+  (if term
+    (nth term 0)
+    []))
+
+(defn coefficient [term]
+  (if term
+    (nth term 1)
+    0))
 
 ;; ## Monomials
 ;;
@@ -75,7 +82,7 @@
 
 ;; ## Polynomials
 
-(declare evaluate make-constant poly->str)
+(declare evaluate make-constant poly->str poly:zero poly:one)
 
 (deftype Polynomial [arity xs->c]
   v/Value
@@ -443,9 +450,11 @@
 (defn partial-derivatives
   "The sequence of partial derivatives of p with respect to each
   indeterminate"
-  [^Polynomial p]
-  (for [i (range (.-arity p))]
-    (partial-derivative p i)))
+  [p]
+  (if (v/number? p)
+    [1]
+    (for [i (range (.-arity ^Polynomial p))]
+      (partial-derivative p i))))
 
 ;; ## Canonicalizer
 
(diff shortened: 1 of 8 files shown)


src/sicmutils/polynomial.cljc
```diff
@@ -443,9 +450,11 @@
 (defn partial-derivatives
   "The sequence of partial derivatives of p with respect to each
   indeterminate"
-  [^Polynomial p]
-  (for [i (range (.-arity p))]
-    (partial-derivative p i)))
+  [p]
+  (if (v/number? p)
```
sritchie, line 454: there is a bigger change we'll need to make, where numbers are considered polynomials.

src/sicmutils/polynomial/gcd.cljc
```diff
@@ -54,8 +54,12 @@
   "Returns a reducer over the function f which will exit early
   if done? becomes true."
   [done? f]
-  (let [rf #(let [c (f %1 %2)]
-              (if (done? c) (reduced c) c))]
+  (let [rf (fn
```
sritchie, line 57: this lets the reduce function be called with a single argument.

src/sicmutils/polynomial/gcd.cljc
```diff
@@ -258,26 +262,34 @@
...
-      :else (binding [*poly-gcd-bail-out* (maybe-bail-out "polynomial GCD" clock *poly-gcd-time-limit*)]
-              (g/abs
-               (gcd-continuation-chain u v
-                                       with-trivial-constant-gcd-check
-                                       with-optimized-variable-order
-                                       #(inner-gcd 0 %1 %2)))))))
+  ([] 0)
+  ([u] u)
+  ([u v]
+   (cond (v/zero? u)   (g/abs v)
```
sritchie, line 268: I moved these up to catch numbers first.

src/sicmutils/simplify.cljc
```diff
@@ -157,38 +156,32 @@
...
+      (simplify-and-flatten)
+      (rules/sqrt-contract)
+      (square-root-simplifier)
+      (clear-square-roots-of-perfect-squares)
+      (simplify-and-flatten)))
 
 (def simplify-expression
   (simplify-until-stable simplify-expression-1 simplify-and-flatten))
 
-(defn simplify-numerical-expression
```
sritchie, line 179: we no longer need this! Simplify does this by default.

test/sicmutils/calculus/manifold_test.cljc
```diff
@@ -216,16 +216,13 @@
                        (/ 1 (sqrt (+ (expt x 2) (expt y 2) 1))))]
         (rep-roundtrips? S2-gnomonic point rep))
 
-      (comment
```
sritchie, line 219: more tests fixed!

PR #342, opened 2021-04-07
ModInt -> deftype, small cleanups in prep for poly changes

sritchie:
In addition to the notes below, I've added a `halt-at` transducer that I'll use in #341.

From the CHANGELOG:

  - Added `sicmutils.calculus.derivative/D-as-matrix` and
     `sicmutils.matrix/as-matrix`, ported from scmutils.

   - converted `sicmutils.modint.ModInt` to a `deftype`; this allows `ModInt`
     instances to be `=` to non-`ModInt` numbers on the right, if the right side
     is equal to the residue plus any integer multiple of the modulus. `v/=`
     gives us this behavior with numbers on the LEFT too, and `ModInt` on the
     right.

     - This change means that `:i` and `:m` won't return the residue and modulus
       anymore. `sicmutils.modint` gains new `residue` and `modulus` functions to
       access these attributes.

   - The JVM version of sicmutils gains more efficient `gcd` implementations
     for `Integer` and `Long` (in addition to the existing native `BigInteger`
     `gcd`), thanks to our existing Apache Commons-Math dependency.

   - `sicmutils.structure/dual-zero` aliases `compatible-zero` to match the
     scmutils interface. Both are now aliased into `sicmutils.env`.


src/sicmutils/matrix.cljc
```diff
@@ -81,7 +80,7 @@
        Associative
        (assoc [_ k entry] (Matrix. r c (assoc v k entry)))
        (containsKey [_ k] (contains? v k))
-       (entryAt [_ k] (.entryAt v k))
+       (entryAt [_ k] (.entryAt ^Associative v k))
```
sritchie, line 83: reflection warning!

PR #347, opened 2021-04-12
warm diggety: a first crack at geometric algebra ('hestenic.cljc')

vanderoops:
Here's another geometric algebra note for when we get this ported over: https://marctenbosch.com/quaternions/


src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+
+(defn- same-rung? [this otha]
+  (instance? (class this) otha))
+
+(defn- different-rung? [this otha]
+  (not (instance? (class this) otha)))
+
+
+;;
+;; A Gradeling is a bevy of Bladoids of the same grade. It may or may not be
```
sritchie, line 193: what's a bevy of bladoids?

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+                     (reduce (fn [sofar bl]
+                               (str sofar (if (empty? sofar) "" " ") bl))
+                             "" bldds)
+                     "}")))
+
+
+(defn gradeling [grade-or-bladoids]
+  (if
+      (integer? grade-or-bladoids)
+    (Gradeling. grade-or-bladoids ())
```
sritchie, line 224: I think you depend on these being vectors in a few spots, so prefer `[]`.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+(defprotocol IGradeling
+  (bladoids [this])
+  (bladoid-with-basis [this b]))
+
+
+(deftype Gradeling [gr bldds]
+  IGradeling
+  (bladoids [_] bldds)
+  (bladoid-with-basis [this b]
+    (first (filter (fn [el] (= (basis el) b))
```
sritchie, line 207: I wouldn't do it here but just for Clojure fun you could also:

```clojure
(some #(when (= b (basis %)) %)
      (bladoids this))
```

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+                   (bladoids this))))
+
+  IGradable
+  (grade [_] gr)
+
+  Object
+  (toString [_] (str "g" gr "{"
+                     (reduce (fn [sofar bl]
+                               (str sofar (if (empty? sofar) "" " ") bl))
```
sritchie, line 216: `(clojure.string/join " " bldds)` for the guts.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  IMV
+  (gradelings [_] grdlgs)
+  (gradeling-of-grade [this gra]
+    (first (filter (fn [grdl] (= (grade grdl) gra))
+                   (gradelings this))))
+  (grades [_] (map grade grdlgs))
+
+  Object
+  (toString [_] (str "MV["
+                     (reduce (fn [sofar gr]
```
sritchie, line 277: `(clojure.string/join " " grdlgs)` for the guts

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+;; In a strict sense this is a redundant type-entity, since one of these
+;; could always be represented as a Gradeling of grade one. But because
+;; vectors have a vaunted position in GA -- they are rightly the generators
+;; of the whole algebra, y' see-- it's worth representing them explicitly.
+;;
+
+(defprotocol IVect
+  (coefs [this]))
+
+(deftype Vect [cfs]
```
sritchie, line 298: Add metadata too!

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+;;
+;; Vect is, of course, a vector (meaning -- yes! -- a real geometric,
+;; mathematical vector, for gods' sake; not an 'array', per lexical cretinism).
+;; In a strict sense this is a redundant type-entity, since one of these
+;; could always be represented as a Gradeling of grade one. But because
+;; vectors have a vaunted position in GA -- they are rightly the generators
+;; of the whole algebra, y' see-- it's worth representing them explicitly.
+;;
+
+(defprotocol IVect
```
sritchie, line 295: same note as above on protocols vs a bare function with a type check.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  IVect
+  (coefs [_] cfs)
+
+  IGradable
+  (grade [_] 1)
+
+  Object
+  (toString [_] (str "v" cfs)))
+
+(defn vect [& cs] (Vect. cs))
```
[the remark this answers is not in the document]
sritchie: you can add

```clojure
(:refer-clojure :exclude [vector])
```

to your namespace declaration and call this `vector` explicitly.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+    (let [invvy (inv otha)]
+      this))
+  (dot [this otha] this)
+  (wdg [this otha] this)
+  (eq? [this otha]
+    (or (instance? (class this) otha)
+        (and (number? otha) (zero? otha))))
+  (grade-part [this n] this))
+
+(defn zero-element []
```
sritchie, line 349: How about doing

```clojure
(def zero-element (ZeroElement.))
```

So you just make this once?

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+    (Bladoid. 0 []))
+  (downRungify [_] 0)
+
+  IHestenic
+  (scl [this s] this)
+  (sum [this otha] otha)
+  (neg [this] this)
+  (dif [this otha] (neg otha))
+  (prd [this otha] this)
+  (inv [this] (/ 1 0))
```
sritchie, line 338: this is cute, but I'd recommend an explicit exception!

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  (quo [this otha]
+    (let [invvy (inv otha)]
```
sritchie, line 340: I'd recommend an explicit exception if that thing is `zero?` (test it explicitly), otherwise return. Skip the work of actually inverting the other side. I know you're doing this to possibly trigger the exception...

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+      this))
+  (dot [this otha] this)
+  (wdg [this otha] this)
+  (eq? [this otha]
+    (or (instance? (class this) otha)
+        (and (number? otha) (zero? otha))))
```
sritchie, line 346: use sicmutils.value to start getting us ready for cljs support: `v/number?`, `v/zero?`

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+
+(defn zero-element []
+  (ZeroElement.))
+
+
+;;
+;; a shadowy cabal of utility-interior mathy bits.
+;;
+
+(defn- bladoid-mult-bladoid [bll blr]
```
sritchie, line 357: you can call this ` bladoid-*-bladoid` if you want, just noting!

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  (let [basl (basis bll)
+        basr (basis blr)]
+    (Bladoid. (* (coef bll) (coef blr) (order-swap-parity basl basr))
+              (basis-prod basl basr))))
+
+(defn- mv-absorb-gradeling [emvy gr]
```
sritchie, line 363: these short names do make it hard to read... I'm not sure what the abbreviation even means! I get the `mv` part, by `e`, `y`?

src/sicmutils/hestenic.cljc
sritchie, line 357: prefer `l-blade`, `r-blade`, yeah? easier to read... or is that just me?

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+    (MV. (sort-on-grades (conj (gradelings emvy) gr)))))
+
+(defn- mv-absorb-bladoid [emvy bl]
+  (mv-absorb-gradeling emvy (asGradeling bl)))
+
+(defn- mv-absorb-mv [mva mvb]
+  (reduce mv-absorb-gradeling
+          mva (gradelings mvb)))
+
+(defn- bladoid-mult-gradeling [bl gr]
```
sritchie, line 379: you see where the generic multimethods would be helpful :)

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  (reduce (fn [outmv bl_rt] (mv-absorb-bladoid outmv (prd bl bl_rt)))
+          (MV. '())
+          (bladoids gr)))
+
+(defn- gradeling-mult-gradeling [grl grr]
+  (reduce (fn [emvy bldlg] (mv-absorb-mv
+                            emvy
+                            (bladoid-mult-gradeling bldlg grr)))
+          (MV. '())
```
sritchie, line 388: prefer `[]`... or `{}` if you take my dictionary suggestion!

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+          (bladoids grl)))
+
+(defn- mv-mult-gradeling [emvy rt_grdl]
+  (reduce (fn [out_mv lt_grdl]
```
sritchie, line 392: woah, first time I've seen underscores!

src/sicmutils/hestenic.cljc
sritchie: Also I think you discovered that `()` doesn't need a quote...

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+            (mv-absorb-mv
+             out_mv
+             (gradeling-mult-gradeling lt_grdl rt_grdl)))
+          (MV. '())
+          (gradelings emvy)))
+
+(defn- mv-map-biexploded-gradelings-and-sum [funq mvl mvr]
+  (reduce mv-absorb-mv
+          (MV. ())
```
sritchie, line 401: You might consider creating and binding this empty multivector once.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  (accordion-down
+   (mv-map-biexploded-gradelings-and-sum
+    funq mvl mvr)))
+
+
+;;
+;; rung conversions: one, yea, unto another.
+;;
+
+(extend-type Bladoid
```
sritchie, line 416: You can do these inside the type definition too, by the way.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  (asMV [this]
+    this))
+
+
+(extend-type Vect
+  IAsGradeling
+  (asGradeling [this]
+    (Gradeling. 1 (filter (fn [bl] (not= 0 (coef bl)))
+                          (map-indexed
+                           (fn [ind cf] (bladoid cf (vector ind)))
```
sritchie, line 451: Prefer the literal `[ind]` to `(vector ind)`, especially if you change `vect` to `vector` above!

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+        this))))
+
+(extend-type MV
+  IScamperUpAndDownLadder
+  (spankOutNothingness [this]
+    (let [nullplop (map
+                    spankOutNothingness
+                    (gradelings this))
+          spnkd (filter
+                 (fn [grdl] (not (number? grdl)))
```
sritchie, line 514: `v/number?`

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+    (sum this (neg otha)))
+  (prd [this otha]
+    (prd (asMV this) (asMV otha)))
+  (inv [this]
+    (let [sq (dot this this)]
+      (scl this (/ 1.0 sq))))
+  (quo [this otha]
+    (prd this (inv otha)))
+  (dot [this otha]
+    (reduce +
```
sritchie, line 695: no need for the `0`, `reduce` calls the 0-arity of its aggregator function if there is no initializer, and that will return `0`.

sritchie (commented):
Okay, so much good stuff here!!!

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  (bladoids [_] bldds)
+  (bladoid-with-basis [this b]
+    (first (filter (fn [el] (= (basis el) b))
+                   (bladoids this))))
+
+  IGradable
+  (grade [_] gr)
+
+  Object
+  (toString [_] (str "g" gr "{"
```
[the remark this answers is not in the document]
sritchie: Whoops, ignore the `print-dup` thing, you only need `print-method`.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  ;; assumes sortedness; sordidness optional
+  (loop [head (first blds)
+         neck (second blds)
+         trunk (rest blds)]
+    (cond (nil? head)
+          '()
+          (nil? neck)
+          (list head)
+          (same-basis? head neck)
+          (recur (sum head neck) (second trunk) (rest trunk))
```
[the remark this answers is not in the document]
sritchie: If you want to start getting into transducer magic, this is faster....

```clojure
(defn- collapse [blades]
  (sequence
   (comp (partition-by basis)
         (map (partial reduce sum)))
   blades))
```

src/sicmutils/hestenic.cljc
vanderoops: done.

src/sicmutils/hestenic.cljc
vanderoops: any reason this'd be preferred? `some`'s vague and indeterminate connotation (in english) feels like a mmiscue here.

src/sicmutils/hestenic.cljc
vanderoops: sho' 'nuf.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+                             "" bldds)
+                     "}")))
+
+
+(defn gradeling [grade-or-bladoids]
+  (if
+      (integer? grade-or-bladoids)
+    (Gradeling. grade-or-bladoids ())
+    (let [gra (if (empty? grade-or-bladoids)
+                -1
```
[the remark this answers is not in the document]
vanderoops: the intent of this signal (-1, here) is that the Gradeling exists but does not yet contain any Bladoids; the first one 'added' will define the grade foreverafter. thus doesn't feel like a job for a singleton (though let's maybe discuss?).

a reason to have as the sentinel value (which could of course be anything other than a non-negative integer)  -1 is that we use this approach in non-lisp implementations; and in languages for which the type of that 'field' is already defined to be an integer, we don't have the possibility of making such a sentinel be e.g. a symbol...

which implicitly asserts value in some pseudo-isomorphism of implementations across different languages...

src/sicmutils/hestenic.cljc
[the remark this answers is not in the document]
vanderoops: splend.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+         arr rseq]
+    (if (or (empty? arr)
+            (<= val (first arr)))
+      scoo
+      (recur (inc scoo) (rest arr)))))
+
+(defn- order-swap-count [basl basr]
+  (loop [cnt 0
+         lind (dec (count basl))]
+    (if (< lind 0)
```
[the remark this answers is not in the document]
vanderoops: for me, those predicates (`pos?` et al.) pertain meaningfully to interrogating the essence of an immutable thing; whereas here we're doing what amounts to a low-level imperative-programming loop, with temporary index variables that are changing and whose instantaneous state needs to be tested. the `(< x 0)` style feels consonant with that. a matter of taste, of course...

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+      cnt
+      (let [lval (nth basl lind)
+            sc (scoot-count lval basr)]
+        (if (= sc 0)
```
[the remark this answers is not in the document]
vanderoops: keep wishing that clojure had a `nonzero?` predicate!

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+          cnt
+          (recur (+ cnt sc) (dec lind)))))))
+
+(defn- order-swap-parity [basl basr]
```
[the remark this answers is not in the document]
vanderoops: we should talk about this given yesterday's feeling that hestenic should remain a standalone(able) module. how does that assertion square with the important inclination to reuse code?

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  (if (even? (order-swap-count basl basr)) +1 -1))
+
+(defn- sort-on-bases [bldds]
+  (sort (fn [bla blb] (compare (basis bla) (basis blb))) bldds))
```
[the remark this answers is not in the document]
vanderoops: wunderschoen! (did not formerly work, by some vague circumstance of... something, but now does)

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+(defn- sort-on-grades [elmnts]
+  (sort-by grade elmnts))
+
+(defn- same-grade? [this otha]
+  (= (grade this) (grade otha)))
+
+(defn- same-rung? [this otha]
+  (instance? (class this) otha))
+
+(defn- different-rung? [this otha]
```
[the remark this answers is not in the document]
vanderoops: have removed `same-rung?`, since it doesn't get used -- thus eliminating the impropriety of defining `different-rung?` like this by eliminating its possibility.

src/sicmutils/hestenic.cljc
vanderoops: (as discussed yesterday)

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+
+
+;;
+;; A Gradeling is a bevy of Bladoids of the same grade. It may or may not be
+;; itself a blade (the 'not' part becomes a possibility for dimensions above
+;; three, which offers bewildering disappointinments like e01 + e12).
+;;
+
+(defprotocol IGradeling
+  (bladoids [this])
```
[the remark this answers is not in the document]
vanderoops: have an incipient bifurcated-semantics protocol use schema brewing... to be jointly chewed next we're in real time...

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+    (let [gra (if (empty? grade-or-bladoids)
+                -1
+                (grade (first grade-or-bladoids)))]
+      (Gradeling. gra (collapse
+                       (sort
+                        (confirm-grades grade-or-bladoids gra)))))))
+
+(defn- gradeling-absorb-bladoid [grdl bl]
+  (let [gra (grade grdl)]
+    (if (< gra 0)
```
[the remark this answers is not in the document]
vanderoops: ibid.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+      (Gradeling. (grade bl) (list bl))
+      (if (not= gra (grade bl))
```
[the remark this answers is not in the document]
vanderoops: that would've forced the error clause below the healthy clause, no?

or were you suggesting deploying `(not (same-grade? ...`? i'd be game, but we've already 'calculated' the Gradeling's grade (as `gra`, in the surrounding `let`) and use it in a bunch of places. argument is that the typographic recurrence of `gra` throughout the function is a useful cue...

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+        (throw (Exception. (str "can't absorb grade" (grade bl)
```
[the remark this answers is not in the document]
vanderoops: tasty! but, as with the comment... somewhere above, what do we want to in cases like this given the desire to keep _hestenic_ a standalone module?

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+                                " blade into grade " gra
+                                "gradeling...")))
+        (if (bladoid-with-basis grdl (basis bl))
+          (Gradeling. gra (map (fn [blel]
+                                 (if (same-basis? bl blel)
+                                   ;; (sum bl blel)
```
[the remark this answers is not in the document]
vanderoops: think what's shaking out is a practice in which there's a cabal of backstage functions for calculating arithmetic stuff (including intermediate results) that we should keep closed under cabal membership (in the set-theoretic sense)... which means we shouldn't e.g. use `sum` in these `defn-` jobbies. know that there's an explicit transgression of this upstairs in `collapse`...

src/sicmutils/hestenic.cljc
vanderoops: prima! toll! fabelhaft! (schon gemacht)

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+
+  Object
+  (toString [_] (str "MV["
+                     (reduce (fn [sofar gr]
+                               (str sofar (if (empty? sofar) "" " ") gr))
+                             "" grdlgs)
+                     "]")))
+
+(defn mv [gradels]
+  (MV. (sort-on-grades gradels)))
```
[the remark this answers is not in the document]
vanderoops: the comment's _arbitrary_ should probably have read _unconstrained_ or something.

the constituent Gradelings do want to be sorted (or fired through a dictionary, as you've recommended). idea of this constructor is that you can dump a pile of Gradelings in without pre-sorting them. though if the code goes this far, it should probably then also run the Gradeling-rung analogue of `collapse` on the resulting sorted collection...

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+
+
+;;
+;; Vect is, of course, a vector (meaning -- yes! -- a real geometric,
+;; mathematical vector, for gods' sake; not an 'array', per lexical cretinism).
```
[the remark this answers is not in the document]
vanderoops: let's for sure discuss the idea of what constitutes probity in a codebase!

src/sicmutils/hestenic.cljc
vanderoops: aye! to be wrassled to the mat...

src/sicmutils/hestenic.cljc
vanderoops: (have questions for you re: the metadata opportunity here in deftype land...)

src/sicmutils/hestenic.cljc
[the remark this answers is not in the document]
vanderoops: i am making a sizzling-sharp decision not to conflate the idea of a programming language vector with a geometric one; the names must therefore remain unambiguously distinct, without regard to namespace or any other implicit context. for this purpose, i've been lucky that `Vect` is largely untrampled in C++, Smalltalk, etc., and have worked a tattered flag into the thin poor soil accordingly...

src/sicmutils/hestenic.cljc
vanderoops: cool. that properly requires a bit of thought and design.

well, actually, on second breath i'm not sure -- one organism's _cute_ is another's can you think of a more compact depiction of noninvertibility?

anyway, let's jaw!

src/sicmutils/hestenic.cljc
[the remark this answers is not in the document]
vanderoops: think i may not follow... tawk?

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+
+(defprotocol IMV
+  (gradelings [this])
+  (gradeling-of-grade [this gra])
+  (grades [this]))
+
+(deftype MV [grdlgs]
+  IMV
+  (gradelings [_] grdlgs)
+  (gradeling-of-grade [this gra]
```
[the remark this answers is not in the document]
vanderoops: mkay. is the recommendation to do with speed/efficiency? or idiomaticity? or?

src/sicmutils/hestenic.cljc
vanderoops: digging it everywhere all the time!

src/sicmutils/hestenic.cljc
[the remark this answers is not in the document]
vanderoops: mathematically it'd work if it were jumbled. probably work algorithmically (given the current codepile) too. so strictly speaking sortedness is a convenience, but, hey, why not? certainly you don't want a random ordering for human readability...

if in a dictionary, i'm assuming that 'give me all the constituent Gradelings in order' undertakes the sort work again each time?

src/sicmutils/hestenic.cljc
vanderoops: again-again: we need to decide on a posture re: using stuff from sicmutils in the context of wanting hestenic to be standalone.

src/sicmutils/hestenic.cljc
vanderoops: could do.

yeah, that's nice.

done!

src/sicmutils/hestenic.cljc
vanderoops: not only have i made that change here; the `l-arg` / `r-arg` thing is now consistently propagated everywhere! america loves it!

src/sicmutils/hestenic.cljc
vanderoops: it's phonetic!

in cooperation with the comment foregoing, all arguments (and internal `let`-variables) now subscribe to a system that afixes `l-` or `r-` or `acc-` onto the left side of `doid`, `grdl`, or `emvy` according as the argument is the left or right Bladoid-or-Gradeling-or-MV in an addition or multiplicatoin.

src/sicmutils/hestenic.cljc
vanderoops: in this case i fear that i do not! all these functions are 'interior' worker-helper-laborer routines that won't be exposed. it is, rather, the lot of `sum` and `prd` and `dif` (etc.) that will get wired up into the generic multimethods, no?

src/sicmutils/hestenic.cljc
vanderoops: i'm all about leaving behind the lists that clj looks down on in favor of []!

all the species here are now converted thusly.

src/sicmutils/hestenic.cljc
vanderoops: still developing what will converge as my typographic schema for this language. suspect underscores will appear somewhere to denote something... but no longer here.

src/sicmutils/hestenic.cljc
vanderoops: sarantastic! done.

hm. looking at it in a bunch of code revealed that `empty-mv` wasn't as visually distinguished as `(MV. [])`... but problem solved by upgrading the name to `the-empty-mv`, which also de-ambiguifies empty's part of speech. have gone back and done `zero-element` --> `the-zero-element` as well.

and also laid in a fresh monosupply of `the-empty-gradeling` for good measure, and because the mood demands it.

src/sicmutils/hestenic.cljc
vanderoops: aye. though not all protocol implementations can be so done, because of forward reference issues.

this one'd work, i think, but i prefer it in a separate corral for cognitive-clustery goodness.

src/sicmutils/hestenic.cljc
vanderoops: as `vect` ain't going nowheauh...

... can ye kindly explain the preference?

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  IAsMV
+  (asMV [this]
+    (asMV (asBladoid this))))
+
+
+;;
+;; simplification (and complexification) of species
+;;
+
+(extend-type Bladoid
```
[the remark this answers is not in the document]
vanderoops: indeed! for the moment i'm finding the rhythmic return of the code to the leftmost column for each type's implementation block to be visually useful...

src/sicmutils/hestenic.cljc
vanderoops: ibid..dibi

src/sicmutils/hestenic.cljc
vanderoops: right. i kind of liked it originally for its signalling that it's specifically a scalar getting accumulated... but have now taken it out.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+(defn- accordion-down [hestish]
+  (downRungify
+   (spankOutNothingness
+    hestish)))
+
+
+;;
+;; all things Bladoid, don't you know. well, not all. in any event, one of these
+;; is a weighted (scaled) basis element of the GA. does not for the moment
+;; presuppose any particular dimension. also does not afford configurable
```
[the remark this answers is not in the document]
vanderoops: need to make some decisions around the two big pieces that want to be configurable: (1) dimensions, and (2) metric.

in most GA systems for compiled languages, these two choices get specified at compile time and then necessarily persist through the execution of the resulting program. that pretty much means you can't e.g. have a dimension-2 and -3 algebra running side by side in one program.

this here clojure certainly affords the opportunity for less ossified specification of dimension and metric. how exactly we want that stuff to work and feel's a good topic.

(noting here for time's posterior that at present the system is implicitly infinite-dimensional...)

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+;; metrics...
+;;
+
+(defprotocol IBladoid
```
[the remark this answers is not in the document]
vanderoops: going to wrap up all the other topics raised in your kind review and then return to this one (and the reader-invertible textual representation bits)...

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  (coef [this])
+  (basis [this])
+  (square-sign [this])
+  (same-basis? [this otha]))
+
+
+(deftype Bladoid [cf bss]
```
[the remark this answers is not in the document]
vanderoops: a personal proscription is in play here. more on that via a different medium.

have adopted the following scheme: all fields' symbols' names will be the obvious ascii, prepended by `f_`.

so in this case we now have
```
(deftype Bladoid [f_coef f_basis]
  IBladoid
  (coef [_] f_coef)
  (basis [_] f_basis)
  ...
```

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+  IBladoid
+  (coef [_] cf)
+  (basis [_] bss)
+  (square-sign [_]
+    (let [gr (count bss)
+          flormp (/ (* gr (- gr 1)) 2)]
+      (if (even? flormp) +1 -1)))
+  (same-basis? [this otha]
```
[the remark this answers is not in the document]
vanderoops: please?

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,785 @@
...
+          (same-basis? head neck)
+          (recur (sum head neck) (second trunk) (rest trunk))
+          :else
+          (conj (collapse trunk) head))))
+
+(defn- confirm-grades
+  ([blds gra]
+   (if (empty? blds)
+     blds
+     (loop [head (first blds)
```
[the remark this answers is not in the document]
vanderoops: apparently 'exclamation mark' is principally british in use, and 'exclamation point' american? not sure why i've really only ever heard the former...

anyway, have adopted one of those for this function's naming.

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,1084 @@
...
+(def ^:dynamic *hestenic-metric* nil)
+
+(defn metric-for-basis-index [ind]
+  (let [emm (get *hestenic-metric* ind)]
+    (if (nil? emm)  1  emm)))
+
+(defmacro with-metric
+  [tric & formses]
+  `(binding [*hestenic-metric* ~tric]
+     ~(conj formses 'do)))
```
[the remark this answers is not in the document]
vanderoops: excellent; thanks! (and: done.)

src/sicmutils/hestenic.cljc
```diff
@@ -0,0 +1,1084 @@
...
+     (confirm-grades! blds (grade (first blds)))))
+  ([blds gra]  ;; ar/2: with grade specified
+   (if (empty? blds)
+     blds
+     (loop [head (first blds)
+            body (rest blds)]
+       (if (not= gra (grade head))
+         (throw (Exception. (str "incompatible grade: "
+                                 head " but should be " gra)))
+         (if (not (empty? body))
```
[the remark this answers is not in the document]
vanderoops: prithee: wherefore?

src/sicmutils/hestenic.cljc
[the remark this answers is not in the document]
vanderoops: ... as to the three-letter naming schema's crypticicity -- one may note that way way down at the bottom of the source code there are now explicitly semantic expansions for all the IHestenic protocol's denizens. e.g.:

`(def grade-involution gri)`

src/sicmutils/hestenic.cljc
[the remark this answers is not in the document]
MagusMachinae: While predicates mostly seem to be implemented in Java for speed, taking `pos?` as a specific example, those implementations are still doing an inequality check at the end of the day, since `pos?` is a wrapper around a function called `isPos`. No prizes for guessing how that's implemented.  A rose by any other name...

The important idea here is binding that procedure to a name that scans faster, clearly communicates the aglorithm being used, and has less arguments you have to worry about getting wrong. I think it only feels consonant because the expectation that looping conditions look like that has been burnt into our minds by decades of C-overexposure.

huahaiy:
This PR appears to be stalled. What's the holdup? Looking forward to using this to learn GA.

vanderoops:
Good day! And thanks for your interest in this GA implementation.

The 'hestenic' chunk of code in question needs a bunch of additional work before it would be appropriate to merge with the scimutils bloodline. And while my ardor for that project remains undimmed, I won't have time for a little while.

It's worth mentioning, though, that the design intent is to provide a set of GA tools that interoperates with the sicmutils system-of-systems but that also functions in a self-contained, standalone manner. And so... if your intent was to play around with GA in the context of sicmutils, then the news is not great short term (though eventually good);  but if you wanted to experiment with 'hestenic' by itself, I could certainly provide access to it in an isolated repo.

					john

sritchie:
@huahaiy this was a boutique exploration into GA, as @vanderoops notes; if you're interested in learning GA with sicmutils, I encourage you to go for it!

What sort of materials are you looking for? An implementation of the types, sure, but beyond that, is there a book you're following along with, or something like that? Or are you looking for essays etc that go through the concepts, in addition to code?

huahaiy:
@vanderoops @sritchie Thanks for responding. I am not familiar with sicmutils at this moment, and certainly would be willing to dive in more, particularly if it has a good GA implementation, as that was my main interest. 

I have a few elementary books on GA, and they all mention the use of some software for some parts of the study. For example, Alan Macdonald's Linear and Geometric Algebra uses a GA module in SymPy; Miroslav Josipovic's Geometric Multiplication of Vectors uses Cl3 in Mathematica; Dorst et al.'s Geomtric algrebra for computer science discusses implementation of GA and has their own C++ implementation. 

I would ideally like an implementation of GA in Clojure to play with, hence my interest in this PR.

src/sicmutils/hestenic.cljc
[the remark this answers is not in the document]
jackrusher: > apparently 'exclamation mark' is principally british in use, and 'exclamation point' american? not sure why i've really only ever heard the former...

... but in our transnational culture, we know it as 'bang'.

PR #348, opened 2021-04-14
Change Complex representation, add new permutation-parity arity

sritchie:
From the CHANGELOG:

- #348:

  - Adds a new single arity version of
    `sicmutils.util.permute/permutation-parity`, which returns the parity of a
    permutation relative to its sorted version.

  - `sicmutils.complex/complex` can now take a single string argument in both
    Clojure and Clojurescript.

  - Expands the complex number literal parser to take these forms, in addition
    to the previously-supported string argument:

```clj
#sicm/complex [1.2 3.6]    ;; 1.2+3.6i
#sicm/complex [1.2]        ;; 1.2
#sicm/complex 1.4          ;; 1.4
#sicm/complex "1.2 + 3.6i" ;; 1.2+3.6i
```


src/sicmutils/util/def.cljc
```diff
@@ -147,32 +147,31 @@
   compiler does not currently error in case 2 and already handles emitting the
   warning for us.)"
   [ns]
-  (fork
```
sritchie, line 150: `fork` did not seem to be doing what I thought it was!

PR #349, opened 2021-04-16
Modernize pattern matching system

sritchie:
From the CHANGELOG:

- #349 introduces a new pattern matching system, built out of matcher
  combinators. All of the rules in `sicmutils.simplify.rules` now use the new
  syntax offered by the library. Some notes:

  - `pattern.match` defines a number of "matcher combinators"; these are
    functions that take a map of bindings, a data input and a success
    continuation and either succeed by calling their continuation, or fail. Out
    of the box, the library provides `fail`, `pass`, `with-frame`,
    `update-frame`, `predicate`, `frame-predicate`, `eq`, `bind`, `match-when`,
    `match-if`, `or`, `and`, `not`, `segment` and `sequence`.

  - Additionally, any combinator that takes another combinator can ALSO take a
    pattern form like `'?x`. See `pattern.syntax` for the full, rich range of
    syntax allowed. These are all functions, so you'll have to quote your
    symbols at this stage.

  - Passing a matcher combinator to `pattern.match/matcher` to generate a
    matcher object. This is a function from some `data` input to a map of
    bindings on success, or an explicit `pattern.match/failure` object on
    failure. Test for failure with `pattern.match/failed?`.

  - A combination of a matcher and a "consequence function" is called a "rule".
    A consequence is a function that takes a binding map and either returns a
    new result or fails by returning `nil` or `false`. (Don't worry, you can
    succeed with these values too by wrapping them in `sicmutils.rule/succeed`.)

    Rules are the heart of the whole simplification mechanism in sicmutils! To
    learn about how to build these, see the documentation for `pattern*`,
    `pattern`, `consequence`, `template`, `rule*`and `rule`.

  - `pattern.rule` gives you some starter rules, and many combinators you can
    use to build more and more powerful and complex sets of rules. These are
    `pass`, `fail`, `predicate`, `return`, `branch`, `choice*`, `choice`,
    `pipe*`, `pipe`, `n-times`, `attempt`, `guard`, `iterated`, `while`,
    `until`, `fixed-point` and `trace`.

  - Rules are nice for rewriting entire expressions recursively, from the bottom
    up or top down. This is called "term rewriting". A big motivation for this
    rewrite was to make it easy to build custom term rewriters for types like
    abstract matrices or abstract up and down structures. You can use your rules
    to rewrite structures recursively with `bottom-up`, `top-down`,
    `iterated-bottom-up` and `iterated-top-down`. `ruleset*`, `ruleset`,
    `rule-simplifier` and `term-rewriting` capture some common patterns the
    library uses to go from rules => term rewriters.

  - If you want ideas about how to use the pattern matching library to rewrite
    expressions, see `sicmutils.simplify.rules` for many examples.


src/pattern/dispatch.cljc
```diff
@@ -0,0 +1,240 @@
...
+;; This software is distributed in the hope that it will be useful, but
+;; WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;; General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this code; if not, see <http://www.gnu.org/licenses/>.
+;;
+
+(ns pattern.dispatch
```
sritchie, line 20: TODO - delete this - make a new version that doesn't make its own type! that is silly, we don't need to append rules.

PR #351, opened 2021-04-20
Change docstring of Euler–Mascheroni constant

jonasemueller:
The docstring for the Euler–Mascheroni constant was describing Euler’s number instead of the Euler–Mascheroni constant.

Alternatively, one can fix this by adjusting the number to be Euler's number. I've created a merge request for that [here](https://github.com/sicmutils/sicmutils/pull/352).

One of the two merge requests should be rejected depending on which solution is preferred.


sritchie:
I love these.. let's take them both! I've edited this PR to include both constants, with your suggestion about euler-gamma taken. Thank you!

PR #352, opened 2021-04-20
Change Euler’s constant to Euler’s number

jonasemueller:
The docstring for `euler` describes Euler’s number while the value was the Euler–Mascheroni constant, also known as Euler’s constant. This changes the value to Euler’s number.

Alternatively, one can fix this by adjusting the docstring to describe the Euler–Mascheroni constant. I've created a merge request for that [here](https://github.com/sicmutils/sicmutils/pull/351).

One of the two merge requests should be rejected depending on which solution is preferred.


sritchie:
Closing, since I've merged this into the other PR. Thank you!!

PR #381, opened 2021-10-31
Add ch7, covariant tests

sritchie:
This PR addresses issues noticed by @phasetr in ch7 of FDG. I also noticed a number of tests in `covariant_derivative.scm` that I had missed; porting those exposed a bug in the text, which is fixed as of this PR.

From the CHANGELOG:

  - `sicmutils.calculus.coordinate/generate` moves to
    `sicmutils.calculus.manifold/c:generate`; this supports a bugfix where
    1-dimensional manifolds like `R1-rect`, aka `the-real-line`, return a
    coordinate prototype of a single element like `t` instead of a structure
    with a single entry, like `(up t)`. Thanks to @phasetr for the bug report
    that led to this fix, and @gjs for finding and fixing the bug.

  - `same.ish/Approximate` implemented for `sicmutils.structure/Structure`,
    allowing `ish?` comparison of `up` and `down` structures with approximate
    entries. Require `sicmutils.generator` for this feature. (NOTE: because
    protocols are implemented for the LEFT argument, `(ish? <vector> (down
    ...))` will still return true if the values are approximately equal, even
    though a `<vector>` is technically an `up` and should NOT equal a `down`. Do
    an explicit conversion to `up` using `sicmutils.structure/vector->up` if
    this distinction is important.)

  - `same.ish/Approximate` now defers to `sicmutils.value/=` for equality
    between `Symbol` and other types. This lets `ish?` handle equality between
    symbols like `'x` and literal expressions that happen to wrap a single
    symbol.

  - `Cartan->Cartan-over-map` now does NOT compose `(differential map)` with its
    internal Cartan forms. This fixed a bug in a code listing in section 7.3 of
    FDG.

  - Section 7.3 of FDG implemented as tests in `sicmutils.fdg.ch7-test`.

  - Many new tests and explorations ported over from `covariant-derivative.scm`.
    These live in `sicmutils.calculus.covariant-test`.

  - timeout exceptions resulting from full GCD are now caught in tests using
    `sicmutils.simplify/hermetic-simplify-fixture`. Previously, setting a low
    timeout where simplification failed would catch and move on in normal work,
    but fail in tests where fixtures were applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@
 - #379 fixes typos in a couple of the equations in `richardson.cljc`, closing
   #377. Thanks to @leifp for the report.
 
+- new PR:
+
+  - approximate equality of up, down implemented. Note trickiness.
+
+  - bugfix in `Cartan->Cartan-over-map` so that the book works now.
+
+  - new tests ported over!
+
+  - Added remaining code forms for chapter 7 in `ch7_test.cljc`
+
 ## 0.19.2
 
 Yet another incremental release, this time to bump the `Fraction.js` dependency.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
 
   - Added remaining code forms for chapter 7 in `ch7_test.cljc`
 
+  - TODO tests for symbol equality in `ish?`
+
 ## 0.19.2
 
 Yet another incremental release, this time to bump the `Fraction.js` dependency.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,19 +10,42 @@
 - #379 fixes typos in a couple of the equations in `richardson.cljc`, closing
   #377. Thanks to @leifp for the report.
 
-- new PR:
+- Features, tests and bugfixes from #381:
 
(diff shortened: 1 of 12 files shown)


sritchie:
@phasetr, I will write more later, but I spoke with Sussman and you found a legit bug in the code from the book!

The fix is in the definition of `g`. Basically `((differential gamma) d:dt)` should just be `d:dt`, since `(differential gamma)` is already applied to the `d:dt` argument inside of`Cartan->Cartan-over-map`:

https://github.com/sicmutils/sicmutils/blob/master/src/sicmutils/calculus/covariant.cljc#L228

Same fix for the scheme code in the book. Here is the correct `g`, and the beginnings of some tests that lock this in:

https://github.com/sicmutils/sicmutils/pull/381/files#diff-6731d08ad0eb5e693e47f0bebee3b1095fee2b86b99ff1606500aea548b53bbeR99-R103

sritchie:
@phasetr ACTUALLY after another round with Sussman, and now that I understand the problem better, this PR will fix the library so that the code from the book works as is. That conceptually makes more sense. Hold tight until I get more tests ported and get this merged!

PR #384, opened 2021-11-05
FDG chapter 9, fix manifold and field equation bug

sritchie:
This PR ports all code listings from chapter 9 of FDG over to tests, and gets Einstein's Field Equations working! I had a super tough bugfix here. I sent GJS an email that I will paste in as a response to this PR. But IT ALL WORKS!!

From the CHANGELOG:

- #384:

  - Adds `sicmutils.fdg.ch9-test`, with tests for all forms from FDG's 9th
    chapter.

  - Tests from `sicmutils.fdg.einstein-test` now all work, and quite fast. The
    functions in this namespace comprise some of the exercises from FDG chapter
    9. (Einstein's Field Equations hung until this PR... getting these working
    is a huge achievement for me, and, in some sense, the final milestone of the
    Big Port from scmutils.)

  - Adds `sicmutils.function/memoize`, a metadata-and-function-arity preserving
    version of `clojure.core/memoize`.

  - in `sicmutils.calculus.indexed`, `with-argument-types` and
    `with-index-types` now both correctly set the arity of the returned
    function, in addition to the argument types or indices.
    `sicmutils.function/arity` will now work correctly with indexed or typed
    functions.

  - Adds new `manifold?` and `manifold-family?` functions in `sicmutils.env` and
    `sicmutils.calculus.manifold`. These are enabled by new `:type
    :sicmutils.calculus.manifold/{manifold,manifold-family}` keys in the
    appropriate structures in the manifold namespace. Manifolds and manifold
    families will now respond with these keywords to `sicmutils.value/kind`.

  - The `sicmutils.calculus.manifold/ICoordinateSystem` now has a `uuid`
    function, for internal comparison of coordinate systems. This is here so
    that points can cache coordinate system representations by UUID. Before this
    change, changing the coordinate prototype, or attaching metadata to a
    coordinate system would break its cache entry in manifold points. (This was
    the killer for the Einstein Field Equations!)

  - `sicmutils.calculus.manifold/{coordinate-prototype,with-coordinate-prototype}`
     now store and retrieve the coordinate prototype from metadata. This plus
     the previous change allows manifold points to correctly cache their
     coordinate representations.

  - `sicmutils.calculus.manifold/manifold` acts as identity on manifolds now.
    Previously it only worked on coordinate systems.


sritchie:
(Here's my email describing the big bug and its fix. I was writing to GJS so translated everything to scheme and scmutils files.)

Hey, moving to this thread... I managed to get my bug fixed! The upshot is:

- the first two examples now take ~90 seconds
- Conservation of energy-momentum takes ~13 seconds for all 4 indices
- the final set of 4 examples take 3.5 seconds total.

Pretty quick! I'm sure this is benefitting from some memoization I added during investigation. That would be nice to talk about; it saves a bunch of duplicate calls, but I suspect that it is wiser to try and reduce the duplicate calls at the source. Let me know if you want to chat, or if you would like me to write up where I was able to save time.

If you are curious what the solution was, the problem was quite subtle:

In `manifold.scm`, manifold points keep a mapping of coordinate system => representation, accessible by COORDINATE-REPS. If you create a point from some manifold like SPACETIME-SPHERICAL using, say, (point spacetime-spherical), you will get a manifold point, BUT that point will have cleverly stashed a clean representation like

```clj
(up t r theta phi)
```

in the COORDINATE-REPS mapping. The representation of the point "as it actually is on the manifold" is in rectangular 4D coordinates:

```clj
(up t
    (* r (sin theta) (cos phi))
    (* r (sin theta) (sin phi))
    (* r (cos theta)))
```

In fact, if you ever try and use `(spacetime-spherical 'point->coords)` to go back, you will NOT get a nice clean (up r t theta phi); you will get:

```clj
(up t
    (sqrt (+ (expt (* r (sin theta) (cos phi)) 2)
             (expt (* r (sin theta) (sin phi)) 2)
             (expt (* r (cos theta)) 2)))
    (acos (/ (* r (cos theta))
             (sqrt (+ (expt (* r (sin theta) (cos phi)) 2)
                      (expt (* r (sin theta) (sin phi)) 2)
                      (expt (* r (cos theta)) 2)))))
    (atan (* r (sin theta) (sin phi))
          (* r (sin theta) (cos phi))))
```

But that does not matter (in scmutils) because MAKE-MANIFOLD-POINT stashes a mapping of spacetime-spherical => `(up r t theta phi)` for the point on creation, case closed!

In my port, I (cleverly, I thought) removed almost all mutation, and changed SET-COORDINATE-PROTOTYPE! to WITH-COORDINATE-PROTOTYPE, which takes an immutable coordinate system and prototype and returns a NEW one.

But this breaks the COORDINATE-REPS cache, since now my coordinate system does not match the old one.

What does WITH-COORDINATE-PROTOTYPE have to do with Einstein's Field Equations??

Well, my tests were using my equivalent of USING-COORDINATES; but I structured my tests so there were TWO separate wrappings of USING-COORDINATES. Because of this, my field equations code was working with points that reported their spacetime-spherical coordinate representation as that crazy mess above, instead of a reasonable

```clj
(up t r theta phi)
```
The simplifier was not pleased.

PR #386, opened 2021-11-12
Tests for FDG chapters 10, 11

sritchie:
This PR:

- adds all code listings from chapters 10 and 11 of FDG as tests.
- tidies all tests and notes from all FDG chapters


sritchie:
@phasetr, this is the final block of tests. Hopefully this will help, and compare favorably to the style you are using on your side!

PR #391, opened 2021-11-14
Replace numsymb with pattern matching versions [in progress]

adamhaber:
Addresses sicmutils/placeholder#70 . Currently, the pattern-matching versions of `numsymb` live in a different namespace `numsymb-rules`. Only `tan`, `atan`, `gcd` and `not` are currently implemented - thought it would be good to ask for feedback in an initial stage to make sure I'm doing this right. 

Very basic testing throughout the code, presumably these should go somewhere in the `/test/sicmutils` folder?


src/sicmutils/numsymb-rules
```diff
@@ -0,0 +1,164 @@
+;;
```
sritchie, line 1: this file needs to be named `numsymb_rules.cljc`. I'd recommend `sicmutils/numsymb/rules.cljc`, actually... use the folder structure to your advantage!

src/sicmutils/numsymb-rules
```diff
@@ -0,0 +1,164 @@
...
+            [sicmutils.numbers]
+            [sicmutils.ratio]
+            [sicmutils.value :as v]
+            [sicmutils.util :as u]
+            [sicmutils.util.aggregate :as ua]
+            [pattern.rule :as r]
+            [sicmutils.numsymb :as sym]))
+
+
+;; Q - what would be the best way to reuse these?
```
sritchie, line 37: best way is to make them non-private and import them.

src/sicmutils/numsymb-rules
```diff
@@ -0,0 +1,164 @@
...
+(def ^:no-doc -pi-over-4-mod-pi? #{'-pi-over-4})
+(def ^:no-doc pi-over-4-mod-pi? #{'pi-over-4 '+pi-over-4})
+
+
+(defn sym:not
+  "For symbolic `x`, returns a symbolic expression representing the logical
+  negation of `x`. For boolean `x`, returns the negation of `x`."
+  [x]
+  (let [RS (r/ruleset
+            (? ?x true?) r/=> false
```
sritchie, line 84: I would import `=>` directly so you don't need the slash. If you are matching a literal you can do `true => false`, or you can write `(? ?x boolean?) => (: #(not (% '?x)))`, I believe.

src/sicmutils/numsymb-rules
```diff
@@ -0,0 +1,164 @@
...
+            (? ?x false?) r/=> true
+            ?x r/=> (not ?x))]
+    (RS x)))
+
+;; basic tests - these should probavbly live elsewhere?
```
sritchie, line 89: see the convention for other namespaces. Put these in `tests`, same namespace with `-test` appended.

src/sicmutils/numsymb-rules
```diff
@@ -0,0 +1,164 @@
...
+(let [v [true false 1 0 'a '() ()]]
```
sritchie, line 90:
```clj
(doseq [v ...]
  (is (= (sym:not v) ((sym/symbolic-operator 'not) v))))
```

src/sicmutils/numsymb-rules
```diff
@@ -0,0 +1,164 @@
...
+(defn tan
+  [x]
+  (let [tan (r/ruleset
+             (? ?x v/zero?) r/=> 0
+             (? ?x v/exact?) r/=> (tan ?x)
+             (? ?x n:zero-mod-pi?) r/=> 0
+             (? ?x n:pi-over-4-mod-pi?) r/=> 1
+             (? ?x n:-pi-over-4-mod-pi?) r/=> -1
+             (? ?x n:pi-over-2-mod-pi?) r/=> (u/illegal "Undefined: tan")
+             (? ?x v/number?) r/=> (?? #(Math/tan (% '?x)))
```
sritchie, line 118: single `?` , not `??`, since you are not splicing in a result

src/sicmutils/numsymb-rules
```diff
@@ -0,0 +1,164 @@
...
+
+
+(let [v [0 1 2 3 4 'a true +]]
+  (= (map #((ua/monoid sym:gcd 0) 2 %) v)
+     (map #((sym/symbolic-operator 'gcd) 2 %) v)))
+
+(defn tan
+  [x]
+  (let [tan (r/ruleset
+             (? ?x v/zero?) r/=> 0
```
sritchie, line 112: use `_` to skip binding (instead of `?x`) if you are not going to use the binding

sritchie (changes requested):
Nice start!

src/sicmutils/numsymb-rules
adamhaber: do you mean `(? #(not (% '?x)))` (? instead of :) ?

src/sicmutils/numsymb-rules
adamhaber: I'm a little confused by the difference; is ?? used for binding collections, whereas ? is for binding "single" variables?

PR #394, opened 2021-11-19
Fix type bug in covariant.cljc tests, bug w/ derivatives of fns returning maps

sritchie:
This PR fixes an issue I encountered when converting Lie.scm in the scmutils codebase. One of the examples errored for me instead of returning the expected value.

Figuring this out helped me tidy up an edge case with derivatives of functions that return maps. I'll add my email to GJS with the scmutils problem details as an issue comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@
 
 ## 0.20.0
 
+- #394 fixes a bug with derivatives of functions that returned a map... but
+  where the map was actually meant to represent some other type, by holding a
+  `:type` key. We do this for manifold families and manifold points, as two
+  examples. Now, instead of recursing into the values, the system will correctly
+  throw an error. (You can fix this by using a `defrecord` instead of a map and
+  implementing `sicmutils.differential/IPerturbed`.)
+
 - #393:
 
   - Forms like `(let-coordinates [(up x y) R2-rect] ...)` will now work even if
--- a/src/sicmutils/collection.cljc
+++ b/src/sicmutils/collection.cljc
@@ -196,7 +196,12 @@
    d/IPerturbed
    (perturbed? [m] (boolean (some d/perturbed? (vals m))))
    (replace-tag [m old new] (u/map-vals #(d/replace-tag % old new) m))
-   (extract-tangent [m tag] (u/map-vals #(d/extract-tangent % tag) m))))
+   (extract-tangent [m tag]
+     (if-let [t (:type m)]
+       ;; Do NOT attempt to recurse into the values if this map is being used as a
+       ;; simple representation for some other type, like a manifold point.
+       (u/unsupported (str "`extract-tangent` not supported for type " t "."))
+       (u/map-vals #(d/extract-tangent % tag) m)))))
 
 
 ;; ## Sets
--- a/test/sicmutils/calculus/covariant_test.cljc
+++ b/test/sicmutils/calculus/covariant_test.cljc
@@ -1019,10 +1019,3 @@
                   gamma)
                  u)
                 ((point the-real-line) 't)))))))))
-
-
-
-;; AHHHHH, so this is hitting the default of extract-tangent!! Tell GJS. You
-;; cannot do that, you probably want a default that errors. In this case, you
-;; are hitting it because you have a derivative of a function that returns a
-;; manifold point.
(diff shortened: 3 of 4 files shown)


sritchie:
```scm
;; GJS,
;;
;; When I was converting Lie.scm, I noticed that the discussion starting at line
;; 168, with "Let phi_t(x) be the integral curve of V from x for interval t",
;; triggered a bug in my Clojure port. When I ran the equivalent of the
;; following code in my system:

(install-coordinates R2-rect (up 'x 'y))

(define R2-rect-point ((R2-rect '->point) (up 'x0 'y0)))

(define X (literal-vector-field 'X R2-rect))
(define Y (literal-vector-field 'Y R2-rect))
(define f (literal-scalar-field 'f R2-rect))

(define ((((Lie-test V) Y) f) x)
  (define (g t)
    (- ((Y f) ((+ identity (* t (V identity))) x))
       ((Y (compose f (+ identity (* t (V identity))))) x)))
  ((D g) 0))


(pec (- ((((Lie-test X) Y) f) R2-rect-point)
     ((((Lie-derivative X) Y) f) R2-rect-point)))
#| Result:
0
|#

;; In the Clojure version, instead of 0, I received an error that "extract-tangent"
;; was not defined on a manifold point.
;;
;; What is more, when I ran the example in scmutils, I did not get 0 either! In
;; the current codebase,

((((Lie-test X) Y) f) R2-rect-point)

;; evaluates to 0, so what is printed is actually equivalent to 
;; (- ((((Lie-derivative X) Y) f) R2-rect-point)):

(pec ((((Lie-test X) Y) f) R2-rect-point)
     (- ((((Lie-derivative X) Y) f) R2-rect-point)))
#| Result:
(+ (* (((partial 1) X^0) (up x0 y0)) (Y^1 (up x0 y0)) (((partial 0) f) (up x0 y0)))
   (* (Y^1 (up x0 y0)) (((partial 1) X^1) (up x0 y0)) (((partial 1) f) (up x0 y0)))
   (* -1 (((partial 1) Y^0) (up x0 y0)) (((partial 0) f) (up x0 y0)) (X^1 (up x0 y0)))
   (* -1 (((partial 0) Y^1) (up x0 y0)) (((partial 1) f) (up x0 y0)) (X^0 (up x0 y0)))
   (* -1 (((partial 1) Y^1) (up x0 y0)) (X^1 (up x0 y0)) (((partial 1) f) (up x0 y0)))
   (* (((partial 0) X^0) (up x0 y0)) (((partial 0) f) (up x0 y0)) (Y^0 (up x0 y0)))
   (* -1 (((partial 0) f) (up x0 y0)) (X^0 (up x0 y0)) (((partial 0) Y^0) (up x0 y0)))
   (* (((partial 0) X^1) (up x0 y0)) (((partial 1) f) (up x0 y0)) (Y^0 (up x0 y0))))
|#

;; Well, I figured out what is going on, and I think the fact that it didn't
;; error exposes an interesting bug.
;;
;; You noted a clue to the issue on line 202:
;;
;; "this result is a consequence of confusing manifold points with tuples of
;; coordinates in the embedding space."
;;
;; Here is how to fix Lie-test on line 189 so that the types work out and the
;; example succeeds again. Notice the redefinition of IDENTITY to be (chart
;; R2-rect), so we get the coordinate representation out:

(define ((((Lie-test V) Y) f) x)
  (let ((I (chart R2-rect)))
    (define (g t)
      (- ((compose (Y f) (point R2-rect))
          ((+ I (* t (V I))) x))
         ((Y (compose f
                      (point R2-rect)
                      (+ I (* t (V I)))))
          x)))
    ((D g) 0)))

(pec (- ((((Lie-test X) Y) f) R2-rect-point)
     ((((Lie-derivative X) Y) f) R2-rect-point)))
#| Result:
0
|#

;; So why did the Clojure port error, but scmutils succeeds with a 0 for the
;; Lie-test term?
;;
;; Lie-test contains this term, which evaluates to 0 in scmutils:

((X identity) R2-rect-point)
;;=> 0

;; But that function should in fact error; (X identity) is supposed to take a
;; real-valued function of the manifold, and IDENTITY here will return its
;; input, ie, a manifold point.
;;
;; The reason this succeeds and evaluates to 0 is that EXTRACT-DX-PART returns 0
;; when passed a manifold point instead of erroring. It probably should error,
;; because it is not in any coordinate representation.

(extract-dx-part 1 R2-rect-point)
;; => 0

;; Here is where this happens, in kernel/diff.scm, lines 819-831:

(define (extract-dx-differential dx obj)
  (if (differential? obj)
      <implementation-elided>
      :zero))

;; So this is a bug, I would say, that is NOT caught because of the default
;; behavior of EXTRACT-DX-PART.
;;
;; The 0 then ripples through Lie-test, masking the type error.
```

PR #435, opened 2021-12-03
Render with SICMUtils namespaces with clerk

ozimos:
add [clerk](https://github.com/nextjournal/clerk) to development dependencies

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,4 +1,5 @@
-{:deps {org.clojure/clojure {:mvn/version "1.10.1"},
+{:paths ["src"]
+ :deps {org.clojure/clojure {:mvn/version "1.10.3"},
         dm3/stopwatch {:mvn/version "0.1.1"},
         cljsjs/odex {:mvn/version "2.0.4-0"},
         org.apache.commons/commons-math3 {:mvn/version "3.6.1"},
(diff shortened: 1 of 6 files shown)


deps.edn
```diff
@@ -10,4 +11,6 @@
         org.clojure/core.match {:mvn/version "1.0.0"},
         cljsjs/complex {:mvn/version "2.0.11-0"},
         borkdude/sci {:mvn/version "0.2.7"},
-        cljsjs/bigfraction {:mvn/version "4.1.1-0"}}}
+        cljsjs/bigfraction {:mvn/version "4.1.1-0"}}
+ :aliases {:dev {:extra-paths ["dev"]
```
sritchie, line 15: I like this, @ozimos , but I think we need to add it to `:deps` proper, since folks will need clerk on the classpath to evaluate `sicmutils.differential`. I think I am fine with that, but let me give this PR a good stare first.

ozimos: @sritchie My thinking is that people who want to read the project source are probably looking to contribute. They'll probably have to pull in the dev alias so it's not an extra burden. Especially if the README includes a mention that this is an easy text editor agnostic way to view the namespace comments. But if SICMUtils is just a library to some other repository, then there's no need to force the repo to pull in clerk.
Same reasoning for the Leiningen project.clj

sritchie:
@ozimos , this is great - I totally get it now, and I appreciate you finding a few typos too!

ozimos:
@sritchie I fixed the merge conflicts

sritchie:
Thank you!!

PR #436, opened 2021-12-03
Recreated Pull Request for direct sum and kronecker products.

MagusMachinae:
Still have your requested changes on hand to finish up the PR, but it occurred to me these might be better off as multimethods.

I don't think I had anything to do with line 330-335, so let me know if you took that out of matrix in the intervening period since I first opened the PR and I can take it out of my branch.


sritchie:
Thanks @MagusMachinae ! I will have some time in the next day or so to give this a review.

PR #437, opened 2021-12-06
implement symmetric memoizer for metrics

adamhaber:
An initial attempt at fixing sicmutils/placeholder#22 . Will add the meta-wrapper + comments once this seems to be working properly.

### DIFF
--- a/src/sicmutils/calculus/connection.cljc
+++ b/src/sicmutils/calculus/connection.cljc
@@ -64,7 +64,7 @@
 
 (defn metric->Christoffel-2 [metric basis]
   {:pre [(b/coordinate-basis? basis)]}
-  (let [metric (f/memoize
+  (let [metric (f/memoize-sym-func
                 (comp f/memoize metric))
         gi (metric/invert metric basis)
         vector-basis (b/basis->vector-basis basis)
--- a/src/sicmutils/function.cljc
+++ b/src/sicmutils/function.cljc
@@ -109,6 +109,16 @@
     (with-meta (core-memoize f)
       m)))
 
+(defn memoize-sym-func
+  [f]
+  (let [mem (atom {})]
+    (fn [x y]
+      (if-let [e (find @mem (sort-by v/freeze [x y]))]
+        (val e)
+        (let [ret (apply f [x y])]
+          (swap! mem assoc (sort-by v/freeze [x y]) ret)
+          ret)))))
+
 (defn get
   "For non-functions, acts like [[clojure.core/get]]. For function
   arguments (anything that responds true to [[function?]]), returns


sritchie:
Hey, this is great. @adamhaber , if you did some measuring and this did not make anything faster, maybe we should just skip it. it is adding more complexity, and it would be nice to have a good place where it really matters before we go down this road.

Thoughts?

adamhaber:
Do you have any intuition as to _why_ this doesn't help? I'm wondering if there's an easy way to keep track of all the function calls to `metric`, and count how many times `metric y x` is called after `metric x y` was already computed. This, plus an estimation of how long it takes to compute `metric y x`, should give a reasonable prediction for the expected speedup, right?

adamhaber:
@sritchie some more data which might be relevant: it seems like the symmetric memoizer is caching the results of the following pairs:
```clojure
[d:dx0 d:dx0] [d:dx1 d:dx0] [d:dx2 d:dx0] [d:dx3 d:dx0] [d:dx1 d:dx1] [d:dx2 d:dx1] [d:dx3 d:dx1] [d:dx2 d:dx2] [d:dx3 d:dx2] [d:dx3 d:dx3]
```
and at least for the faster test:

```clojure
(with-literal-functions [R p rho]
    (let [metric (FLRW-metric 'c 'k R)
          basis (e/coordinate-system->basis spacetime-sphere)
          nabla (e/covariant-derivative
                 (e/Christoffel->Cartan
                  (e/metric->Christoffel-2 metric basis)))
          ws    (e/basis->oneform-basis basis)]
      (time (is (= ['(/ (+ (* (expt c 2) ((D rho) t) (R t))
                                 (* 3 (expt c 2) ((D R) t) (rho t))
                                 (* 3 ((D R) t) (p t)))
                              (* (expt c 2) (R t)))
                          0 0 0]
                         (map (fn [i]
                                (simplify
                                 ((e/contract
                                   (fn [ei wi]
                                     (((nabla ei)
                                       (Tperfect-fluid rho p 'c metric))
                                      (nth ws i)
                                      wi))
                                   basis)
                                  ((point spacetime-sphere) (up 't 'r 'theta 'phi)))))
                              (range 4)))))))
```
`metric` is called 131692 times.

So the only missing piece of the puzzle is (I think) the actual time it takes to compute `metric y x` - what would be the best way to time this?

sritchie:
@adamhaber , I say we drop this - if the benchmark shows no difference in times, I don't think it's worth the extra code complexity for now. 

It IS a nice project, if you did want to try doing this in a more general way - write a memoizer that can accept some abstraction of "symmetries", like sorting its arguments, or taking some function to put them in a canonical order.

trace(A*B*C) == trace(B*C*A) == trace(C*A*B) for example.

I'll close, but let me know what you think. Thanks for this investigation!

PR #448, opened 2022-01-04
infinite? and fixes for generic arithmetic ops

sritchie:
From the CHANGELOG:

  - new `g/infinite?` generic with implementations for all numeric types,
    complex numbers, `differential` instances. Defaults to `false` for all other
    types. (Also aliased into `sicmutils.env/infinite?`).

  - The infix, TeX and JavaScript renderers (`->infix`, `->TeX` and
    `->JavaScript`) all properly render `##Inf` and `##-Inf`. Infix uses the
    Unicode symbol ∞, while `->TeX` uses the LaTeX command `\infty`.
    Javascript's `Infinity` stands in for `##Inf` in generated JS code.

  - `g/+`, `g/-`, `g//` now short circuit if there is a NUMERIC zero on either
    side. This was causing bugs in cases where we allow, say, a scalar to be
    added to a quaternion, and auto-convert the scalar right there (so it adds
    only to the real part). OR in cases, like in the matrix PR, where we convert
    the scalar in addition to `<scalar>*I*`.

    - This caused some problems with `sicmutils.matrix` tests that were not well
      typed.

  - The default `expt` implementation is now available as a function to call
    directly (`sicmutils.generic/default-expt`) without going through the
    dispatch system.

### DIFF
--- a/src/sicmutils/generic.cljc
+++ b/src/sicmutils/generic.cljc
@@ -186,12 +186,7 @@
            (and numy? (v/one? y)) x
            :else (mul x y))))
   ([x y & more]
-   (reduce (fn [l r]
-             (if (v/zero? l)
-               (reduced l)
-               (* l r)))
-           (* x y)
-           more)))
+   (reduce * (* x y) more)))
 
 (declare div)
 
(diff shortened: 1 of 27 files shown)


src/sicmutils/generic.cljc
```diff
@@ -347,6 +358,13 @@
 (defmethod negative? :default [a]
   (< a (v/zero-like a)))
 
+(defgeneric infinite? 1
```
sritchie, line 361: still pending.

PR #471, opened 2022-01-23
Complex GCD tests, install generic implementation

sritchie:
- #471:

  - installs the complex GCD implementation into the generic system and modifies
    it to work with real/complex pairs.

  - tweaks the default gcd implementation so that two identical values `x`, even
    if they are floating point, will return `x` from `(gcd x x)`. (Really this
    should check on their absolute values...)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## [unreleased]
 
+- #471:
+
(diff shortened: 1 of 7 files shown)


sritchie:
cc @adamhaber , I moved the GCD implementation to `sicmutils.complex` and tweaked it to work with reals on either side. Good stuff!!

PR #483, opened 2022-02-19
allow pattern matching segment variables to take predicates [in progress]

### LINKED ISSUE #13: Pattern rule segments don't respect predicates
Tried:
```Clojure
(def ruleset-test
  (rule/ruleset
   (+ (? a v/number?) (+ (?? b string?))) =>
   (+ (? a) (?? b))))
(let [rs ruleset-test
      s (rule/rule-simplifier rs)

      thing1 "test1"
      thing2 "test2"]

  (s `(~'+ 5.0 (~'+ ~thing1
                     ~thing2
                    5.0))))
```
Expected:
No change (as second list contains a number)

Observed:
```Clojure
(+ 5.0 "test1" "test2" 5.0)
```
Simplification happens and so predicate is ignored. This has been verified to happen with newly defined types (predicates) and doesn't happen when using non-segment pattern matches.

sritchie:


sritchie:
Covered in Emmy

PR #496, opened 2022-03-15
Enhance compilation API

sritchie:
- #496:

  - replaces the function values in `sicmutils.expression.compile` with symbols;
    I hadn't realized before that substituting in symbolic `Math/sqrt`, for
    example, was possible, vs a `#(Math/sqrt %)` function value. Compiled
    functions are now faster!

    A simulation run of the double pendulum example in the [clerk-demo
    repository](https://github.com/nextjournal/clerk-demo/blob/20a404a271bea29ef98ee4e60a05e54345aa43ba/notebooks/sicmutils.clj)
    now runs in 700ms vs the former 2.2 seconds, a major win.

  - Function compilation now pre-simplifies numerical forms encountered inside a
    function, like `(/ 1 2)`, instead of letting them be evaluated on every fn
    call.

  - All numerical forms encountered in function compilation are now converted to
    either `double` on the JVM or `js/Number` in javascript; this way no
    `BigInt` values etc are left around.

  - In `sicmutils.expression.compile`:

    - gains a new, validating `compiler-mode` function for fetching the compiler
      function.

    - `set-compiler-mode!` now actually works. It never did!

  - New `:source` compile mode that returns a source code form. You can either
    call `eval` on this or call `sci-eval` to get an SCI-evaluated function with
    all proper bindings in place.

  - `compile-state-fn` now takes an optional options map, with support for
    `:flatten?` and `:generic-params?` keywords. These can be used to tune the
    shape of the function returned by `compile-state-fn`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "npm": "6.14.8"
   },
   "devDependencies": {
-    "markdown-it": "12.2.0",
+    "markdown-it": "12.3.2",
     "markdown-it-block-image": "0.0.3",
     "markdown-it-sidenote": "github:gerwitz/markdown-it-sidenote",
     "markdown-it-texmath": "0.9.1",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@
     call `eval` on this or call `sci-eval` to get an SCI-evaluated function with
     all proper bindings in place.
 
+  - `compile-state-fn` now takes an optional options map, with support for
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "npm": "6.14.8"
   },
   "devDependencies": {
-    "markdown-it": "12.2.0",
+    "markdown-it": "12.3.2",
```
sritchie, line 16: this removes the security vuln github was reporting.

PR #522, opened 2022-07-25
Fix link to SICM HTML version

teodorlu:
Looks like the HTML version moved from https://tgvaughan.github.io to https://tgvaughan.github.io/sicm.


sritchie:
Thank you!

PR #523, opened 2022-09-22
Complex number type-hints, fast compiled functions

sritchie:
This PR:

- Adds `^js` type hints to the Complex namespace, similar to the approach in sicmutils/sicmutils#371. This should fix an issue that came up with operations on complex numbers under advanced compilation.
- adds the `cljs` alias into `.dir-locals.el`, fixing an issue where `cider-jack-in-cljs` failed to work in my emacs configuration.
- fixes a typo in the README.
- Working on sicmutils/placeholder#5 

wow, making JS literals is WAY faster than calling aset.

WAYYYYY faster to emit js and eval it if we're on the cljs side, vs using sci.

## Notes

- to get the state without triggering side effects in cljs, atom and ratom use (.-state my-ratom). that should let me delete the goofy thing I was doing to mirror state around in `emmy-viewers`.
- I can delete all `:include-macros true` code if I do the proper 

```
#?(:cljs
     (:include-macros [pattern.rule]))
```

in all cljc namespaces.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -3,4 +3,4 @@
      (cider-default-cljs-repl . node)))
  (clojurec-mode
   . ((cider-preferred-build-tool . clojure-cli)
-     (cider-clojure-cli-aliases . ":test:dev/repl"))))
+     (cider-clojure-cli-aliases . ":test:cljs:dev/repl"))))
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ the Github page. Citation information is generated from
 Here is the generated BibTeX entry:
 
 ```
-@software{Ritchie_SICMUtils_Functional_Computer_2016,
+@software{Ritchie_SICMUtils_Functional_Computer_2016},
 author = {Ritchie, Sam and Smith, Colin},
 license = {GPL-3.0},
 month = {4},
@@ -215,7 +215,6 @@ title = {{SICMUtils: Functional Computer Algebra in Clojure}},
 url = {https://github.com/sicmutils/sicmutils},
 version = {0.22.0},
 year = {2016}
-}
 ```
 
 In the above BibTeX entry, the version number is intended to be that from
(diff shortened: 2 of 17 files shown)


sritchie:
```
(ns demo.bench
  (:require [sci.core :as sci]))

(def sci-context
  (sci/init
   {:classes {'Math js/Math}}))

(defn sci-eval [f-form]
  (sci/eval-form (sci/fork sci-context) f-form))


(let [f1 (fn [G__9 G__10 G__8]
           (let [y3 (aget G__9 1)
                 y4 (aget G__9 2)
                 p5 (aget G__10 0)
                 p7 (aget G__10 2)]
             #js [1 y4 (/ (* -1 p5 (Math/sin y3)) p7)]))
      f2 (sci-eval
          '(fn [G__9 G__10 G__8]
             (let [y3 (aget G__9 1)
                   y4 (aget G__9 2)
                   p5 (aget G__10 0)
                   p7 (aget G__10 2)]
               #js [1 y4 (/ (* -1 p5 (Math/sin y3)) p7)])))
      state  #js [0 3 0]
      params #js [9.8 1 1]
      out (make-array 3)
      n 1000000]
  (js/console.log (time
                   (dotimes [_ n]
                     (f1 state params out))))
  (js/console.log (time
                   (dotimes [_ n]
                     (f2 state params out)))))


;; weird, 6x slower to use the aset above??

;; "Elapsed time: 6.000000 msecs"
;; "Elapsed time: 1817.200000 msecs"

```

sritchie:
Here are some thoughts on function compilation.

I had thought that the core idea was always going to be function to function. But that is not quite right. Yes, we want to be able to call functions of generics, get out an expression body and then wrap it up again! But the important thing is the expression body. 

We want a suite of functions that can go from some expression body to an optimized expression body, probably given some set of symbols that are assumed to be present in lexical scope. And then we can DECIDE once we get the body back what to do. Do we wrap it up in a function and eval? Do we emit some JavaScript?

sritchie:
Superceded by work in Emmy.

